Issue #24, opened 2020-03-22
lib_lightgbm.dylib library missing


ablaom:
```julia
julia> using LightGBM, MLJBase
[ Info: Precompiling MLJBase [a7f614a8-145f-11e9-1d2a-a57a1082229d]

julia> X, y = @load_boston;

julia> model = LightGBM.MLJInterface.LGBMRegression();

julia> MLJBase.fit(model, 1, X, y)
ERROR: error compiling #fit: error compiling #fit#23: error compiling LGBM_DatasetCreateFromMat: could not load library "/Users/anthony/.julia/packages/LightGBM/tA9ec/src/lib_lightgbm.dylib"
dlopen(/Users/anthony/.julia/packages/LightGBM/tA9ec/src/lib_lightgbm.dylib, 1): Library not loaded: /usr/local/opt/libomp/lib/libomp.dylib
  Referenced from: /Users/anthony/.julia/packages/LightGBM/tA9ec/src/lib_lightgbm.dylib
  Reason: image not found
Stacktrace:
 [1] fit(::error in running finalizer: ErrorException("error compiling Booster_finalizer: could not load library "/Users/anthony/.julia/packages/LightGBM/tA9ec/src/lib_lightgbm.dylib"
dlopen(/Users/anthony/.julia/packages/LightGBM/tA9ec/src/lib_lightgbm.dylib, 1): Library not loaded: /usr/local/opt/libomp/lib/libomp.dylib
  Referenced from: /Users/anthony/.julia/packages/LightGBM/tA9ec/src/lib_lightgbm.dylib
  Reason: image not found")
LightGBM.MLJInterface.LGBMRegression, ::Int64, ::NamedTuple{(:Crim, :Zn, :Indus, :NOx, :Rm, :Age, :Dis, :Rad, :Tax, :PTRatio, :Black, :LStat),NTuple{12,Array{Float64,1}}}, ::Array{Float64,1}, ::Array{AbstractFloat,1}) at /Users/anthony/.julia/packages/LightGBM/tA9ec/src/MLJInterface.jl:198
 [2] fit at /Users/anthony/.julia/packages/LightGBM/tA9ec/src/MLJInterface.jl:191 [inlined]
[...]

Issue #1, opened 2020-01-17
Implement LGBM_BoosterSaveModelToString


yalwan-iqvia:
Documentation here: https://lightgbm.readthedocs.io/en/latest/C-API.html#c.LGBM_BoosterSaveModelToString

https://github.com/IQVIA-ML/LightGBM.jl/blob/b5293367b451b9a8fcd78d055890a384bc8c2485/src/wrapper.jl#L505

Why? Currently a booster is passed around as a PTR to memory, which isn't great and is a massive vom for distributed computing of any form, so it should probably be serialised and deserialised. This is the first step towards achieving that

yalwan-iqvia:
#34  closes

PR #31, opened 2020-03-31
Mlj num classes

yalwan-iqvia:
Upgrade to StatsBase

num_class is now auto-filled by MLJ interface and not accepted as a parameter

Dictinction between binary/multiclass is removed (both MLJ and entire codebase)

User can now pass an objective. MLJ interface only checks that an objective is either a valid classification or regression objective.

### DIFF
--- a/src/MLJInterface.jl
+++ b/src/MLJInterface.jl
@@ -12,8 +12,18 @@ const LGBM_METRICS = (
     "xentlambda", "kullback_leibler", "kldiv",
 )
 
+const CLASSIFICATION_OBJECTIVES = (
+    "binary", "multiclass", "softmax", "multiclassova", "multiclass_ova", "ova", "ovr",
+)
 
-MLJModelInterface.@mlj_model mutable struct LGBMRegression <: MLJModelInterface.Deterministic
+const REGRESSION_OBJECTIVES = (
+    "regression", "regression_l2", "l2", "mean_squared_error", "mse", "l2_root", "root_mean_squared_error", "rmse",
+    "regression_l1", "l1", "mean_absolute_error", "mae", "huber", "fair", "poisson", "quantile", "mape",
+    "mean_absolute_percentage_error", "gamma", "tweedie",
+)
+
+
+MLJModelInterface.@mlj_model mutable struct LGBMRegressor <: MLJModelInterface.Deterministic
 
     # Hyperparameters, see https://lightgbm.readthedocs.io/en/latest/Parameters.html for defaults
     num_iterations::Int = 10::(_ >= 0)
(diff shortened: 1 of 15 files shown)


yalwan-iqvia:
@ablaom are you able to test from this?
Please remember that you'll need to take the python binary and change its extension to dylib (I still can't get my head around why this works -- one thing I forgot to check was if the file has execute bit set or not)

yalwan-iqvia:
@ablaom could you do your tests using this branch? If you're satisfied I can make a release.

I have dealt with

- Upgrade to StatsBase
- num_class is now auto-filled by MLJ interface and not accepted as a parameter
- Dictinction between binary/multiclass is removed (both MLJ and entire codebase)
- User can now pass an objective. MLJ interface only checks that an objective is either a valid classification or regression objective.

ablaom:
https://github.com/IQVIA-ML/LightGBM.jl/issues/24#issuecomment-609501457

src/MLJInterface.jl
```diff
@@ -182,142 +135,128 @@ function mlj_to_kwargs(model::MLJModelInterface.Supervised)
...
-    report = LightGBM.fit(model, X, y_lgbm; verbosity=verbosity, weights=w)
+    report = LightGBM.fit!(model, X, y_lgbm; verbosity=verbosity, weights=w)
 
     model = (model, classes)
     cache = nothing
     report = (report,)
 
     # Caution: The model is a pointer to a memory location including its training data
-    # which is probably bum ... fix in LightGBM.jl?
+    # This definitely needs fixing
```
iqml, line 178: Make a separate issue?

src/predict.jl
```diff
@@ -3,7 +3,8 @@
     predict(estimator, X; [predict_type = 0, num_iterations = -1, verbosity = 1,
     is_row_major = false])
 
-Return an array with the labels that the `estimator` predicts for features data `X`.
+Return a **MATRIX** with the labels that the `estimator` predicts for features data `X`.
+Use `dropdims` if a vector is required.
```
iqml, line 7: How about a return type sig just to make this clear?

src/MLJInterface.jl
yalwan-iqvia: https://github.com/IQVIA-ML/LightGBM.jl/issues/1

With a comment from thibaut in there that I seem to have missed..

src/predict.jl
yalwan-iqvia: Ever since i discovered that adding a return type makes julia call convert on the output, I'm a lot less eager to do o, although I'll do it anyway

ablaom: Yes, I think it is generally better to write type-stable code that makes these type annotations on return types unnecessary. Of course comments indicating the type are helpful.

yalwan-iqvia: e99287d

yalwan-iqvia:
@ablaom I've merged this so do let me know if it doesn't quite fit the bill. I won't do a release until you've confirmed satisfied. Please update me if you still are unable to get going due to binary issue

PR #33, opened 2020-04-09
Fix MLJ regression 2d outputs

yalwan-iqvia:
This was deliberate but should not have propogated to MLJ
Added simple full integration test as suggested by @ablaom 
For some reason, classifiction doesn't work so is commented out pending further resolution later


yalwan-iqvia:
@ablaom I've had to disable the tests you suggseted due to problems with outputting

I do like them however, and they pass when run in the REPL. So on this basis I will merge this PR so you can continue your testing and we can figure output problems seperately and I will re-enable tests.

PR #34, opened 2020-04-09
Save model to string

yalwan-iqvia:
I noticed that passing pointers around is multiprocessing and deepcopy unsafe, so this to implement in-memory serialisation

- Implemented `LGBM_BoosterSaveModelToString` C-API
- Implemented `deepcopy_internal` for `LightGBM.Booster` object, which uses the above to serialise and then reload a new booster from the serialised form
- Implemented that fitting will store the serialised model, so that `predict` will load from the serialised model if the booster happens to be C_NULL (which seems likely from usage of serialise for IPC etc, where pointers are always set to C_NULL, andrightly so)
- Added tests
- Updated Github CI matrix to turn off fail-fast

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,6 +9,7 @@ jobs:
     name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }}
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
       matrix:
         version: ['1.0', '1.1', '1.2', '1.3']
         os:  [windows-latest, ubuntu-latest, macOS-latest]
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        version: ['1.0', '1.1', '1.2', '1.3']
+        version: ['1.0', '1.1', '1.2', '1.3', '1.4']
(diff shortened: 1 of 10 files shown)


src/wrapper.jl
```diff
@@ -502,7 +522,39 @@ function LGBM_BoosterSaveModel(bst::Booster , start_iteration::Integer , num_ite
     return nothing
 end
 
-# function LGBM_BoosterSaveModelToString()
+function LGBM_BoosterSaveModelToString(bst::Booster, start_iteration::Integer, num_iteration::Integer)::String
+
+    # places forthe call to write to
```
iqml, line 527: typo

yalwan-iqvia: 4effd1c

.github/workflows/main.yml
```diff
@@ -9,8 +9,9 @@ jobs:
     name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }}
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
       matrix:
-        version: ['1.0', '1.1', '1.2', '1.3']
+        version: ['1.0', '1.1', '1.2', '1.3', '1.4']
```
kd-iqvia, line 14: nice I can see an update on the versions ;)

yalwan-iqvia: They passed too ;)

PR #51, opened 2020-05-14
Mlj update method

yalwan-iqvia:
Currently a WIP to implement the `update` method, effectively allowing to support learning curves over number of iteration. It isn't clear that other parameters can/should be supported for this, but this can be considered if a strong argument can be made for it.

Intention is to make this work by passing number of iterations to the fitting method, such that we could re-call fit with a number of additional iterations to perform on an existing fitted booster.

Have implemented multiple dispatch for num_iterations in fit, next part is to plumb it into MLJ interface.

### DIFF
--- a/src/MLJInterface.jl
+++ b/src/MLJInterface.jl
@@ -170,8 +170,8 @@ function fit(mlj_model::MODELS, verbosity::Int, X, y, w=AbstractFloat[])
     w = Float32.(w)
     report = LightGBM.fit!(model, X, y_lgbm; verbosity=verbosity, weights=w)
 
-    fitresult = (model, classes, mlj_model)
-    cache = nothing
+    fitresult = (model, classes, deepcopy(mlj_model))
+    cache = (num_boostings_done=[LightGBM.get_iter_number(model)], )
     report = (report,)
 
     return (fitresult, cache, report)
@@ -183,17 +183,35 @@ function update(mlj_model::MLJInterface.MODELS, verbosity::Int, fitresult, cache
     old_lgbm_model, old_classes, old_mlj_model = fitresult
 
     # we can continue boosting if and only if num_iterations has changed
-    if MLJModelInterface.is_same_except(old_mlj_model, mlj_model, :num_iterations)
-        additional_iterations = mlj_model.num_iterations - old_mlj_model.num_iterations
-        # eh this is ugly, possibly prompts a need for some refactoring
-        report = LightGBM.train!(old_lgbm_model, additional_iterations, String[], verbosity, LightGBM.Dates.now())
-        # should probably dump report too, its just an empty dict...
-        fitresult = (old_lgbm_model, old_classes, mlj_model)
-        return (fitresult, cache, (report,))
+    if !MLJModelInterface.is_same_except(old_mlj_model, mlj_model, :num_iterations)
+        # if we get here it's just means we need to call fit directly
+        return MLJInterface.fit(mlj_model, verbosity, X, y, w)
     end
 
-    # if we get here it's just means we need to call fit directly
-    return MLJInterface.fit(mlj_model, verbosity, X, y, w)
+    additional_iterations = mlj_model.num_iterations - old_mlj_model.num_iterations
+    if verbosity >= 1
+        @info("Not refitting from scratch", additional_iterations)
+    end
+
+    # splice the data into the estimator -- we need to update num_iterations,
+    # taking into account it may have stopped early previously
+    # It might well stop early again too, but give it a chance
+    # Also ideally early stopping would be implemented via this mechanism and MLJ anyway
+    num_iterations = sum(cache.num_boostings_done)
+    old_lgbm_model.num_iterations = num_iterations + additional_iterations
+
+    # eh this is ugly, possibly prompts a need for some refactoring
+    report = LightGBM.train!(old_lgbm_model, additional_iterations, String[], verbosity, LightGBM.Dates.now())
+
+    # should probably dump report too, its just an empty dict...
+    fitresult = (old_lgbm_model, old_classes, deepcopy(mlj_model))
+
+    final_num_iter = LightGBM.get_iter_number(old_lgbm_model)
+    iteration_history = deepcopy(cache.num_boostings_done)
+    push!(iteration_history, final_num_iter - num_iterations)
+    newcache = (num_boostings_done=iteration_history, )
+
+    return (fitresult, newcache, (report,))
 
 end
 
--- a/src/MLJInterface.jl
+++ b/src/MLJInterface.jl
@@ -189,6 +189,15 @@ function update(mlj_model::MLJInterface.MODELS, verbosity::Int, fitresult, cache
     end
 
     additional_iterations = mlj_model.num_iterations - old_mlj_model.num_iterations
+
+    if additional_iterations < 0
+        # less iterations isn't very valid so re-fit from scratch
+        # TODO: I think theres a LightGBM API where you can prune
+        # the boosting, so we would just do that instead of wasting time re-fitting
+        # although we'd still have to consider fit-report etc
+        return MLJInterface.fit(mlj_model, verbosity, X, y, w)
+    end
+
     if verbosity >= 1
         @info("Not refitting from scratch", additional_iterations)
     end
--- a/src/MLJInterface.jl
+++ b/src/MLJInterface.jl
@@ -157,7 +157,7 @@ function mlj_to_kwargs(model::LGBMClassifier, classes)
 end
 
 # X and y and w must be untyped per MLJ docs
-function fit(mlj_model::MODELS, verbosity::Int, X, y, w=Vector{AbstractFloat}())
+function fit(mlj_model::MODELS, verbosity::Int, X, y, w=AbstractFloat[])
 
     # MLJ docs are clear that 0 means silent. but 0 in LightGBM world means "warnings"
     # and < 0 means fatal logs only, so we put intended silence to -1 (which is probably the closest we get)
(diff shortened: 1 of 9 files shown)


src/fit.jl
```diff
@@ -21,10 +23,13 @@ array that holds the validation metric's value at each iteration.
 * `weights::Vector{Tw<:Real}`: the training weights.
 * `init_score::Vector{Ti<:Real}`: the init scores.
 """
-function fit!(estimator::LGBMEstimator, X::Matrix{TX},
-    y::Vector{Ty}, test::Tuple{Matrix{TX},Vector{Ty}}...; verbosity::Integer = 1,
-    is_row_major = false, weights::Vector{Tw} = Vector{Float32}(),
-    init_score::Vector{Ti} = Vector{Float64}()) where {TX<:Real,Ty<:Real,Tw<:Real,Ti<:Real}
+function fit!(
+    estimator::LGBMEstimator, num_iterations::Int, X::Matrix{TX}, y::Vector{Ty}, test::Tuple{Matrix{TX},Vector{Ty}}...;
```
yalwan-iqvia, line 27: Needs some thought, I am not sure chucking in num_iterations here like this is sensible

If we're updating an existing booster, we don't want to go through the rigmarole of setting up the dataset again

src/utils.jl
```diff
@@ -6,7 +6,7 @@ const BOOSTERPARAMS = [:application, :learning_rate, :num_leaves, :max_depth, :t
                        :feature_fraction, :feature_fraction_seed, :bagging_fraction,
                        :bagging_freq, :bagging_seed, :early_stopping_round, :sigmoid,
                        :is_unbalance, :metric, :is_training_metric, :ndcg_at, :num_machines,
-                       :local_listen_port, :time_out, :machine_list_file, :num_class,:device]
+                       :local_listen_port, :time_out, :machine_list_file, :num_class, :device_type]
```
yalwan-iqvia, line 9: This should have been `device_type` to begin with

yalwan-iqvia:
@ablaom please take a look.

As always, all comments appreciated.

I did not follow `@test_logs` approach here, because I felt like it was incorrect. Instead I have an alternative approach, which I hope you'll approve of, for inspecting numbers of iterations done. You can see the form of testing in `test/mlj/update.jl` which is dedicated to this function. If you think a form of test is missing as well, please do let me know.

@iqml @kd-iqvia @sbeura @mrasack please also take a look and provide any feedback.

src/MLJInterface.jl
```diff
@@ -170,11 +170,30 @@ function fit(mlj_model::MODELS, verbosity::Int, X, y, w=Vector{AbstractFloat}())
...
+end
+
+function update(mlj_model::MLJInterface.MODELS, verbosity::Int, fitresult, cache, X, y, w=AbstractFloat[])
+
+    old_lgbm_model, old_classes, old_mlj_model = fitresult
+
+    # we can continue boosting if and only if num_iterations has changed
+    if MLJModelInterface.is_same_except(old_model, mode, :num_iterations)
+        additional_iterations = mlj_model.num_iterations - old_mlj_model.num_iterations
+        # eh this is ugly, possibly prompts a need for some refactoring
```
ablaom: I'd check this is non-negative and otherwise flip to a cold restart (`fit` call).

src/MLJInterface.jl
```diff
@@ -219,6 +238,11 @@ MLJModelInterface.fit(model::MLJInterface.MODELS, verbosity::Int, X, y) = MLJInt
 MLJModelInterface.fit(model::MLJInterface.MODELS, verbosity::Int, X, y, w::Nothing) = MLJInterface.fit(model, verbosity, X, y)
 MLJModelInterface.fit(model::MLJInterface.MODELS, verbosity::Int, X, y, w) = MLJInterface.fit(model, verbosity, X, y, w)
 
+# Add `update` method
+MLJModelInterface.update(model::MLJInterface.MODELS, verbosity::Int, fitresult, cache, X, y) = MLJInterface.update(model, verbosity, fitresult, cache, X, y)
+MLJModelInterface.update(model::MLJInterface.MODELS, verbosity::Int, fitresult, cache, X, y, w::Nothing) = MLJInterface.update(model, verbosity, fitresult, cache, X, y)
```
ablaom, line 261: On the left-hand side: Rather than `y, w::Nothing)`, this should be `y, w=nothing)`:

https://alan-turing-institute.github.io/MLJ.jl/dev/adding_models_for_general_use/#Summary-of-methods-1

I think otherwise dispatch will fallback to doing a fit instead of update when user provided weights, I reckon.

src/MLJInterface.jl
```diff
@@ -189,7 +189,7 @@ function update(mlj_model::MLJInterface.MODELS, verbosity::Int, fitresult, cache
         report = LightGBM.train!(old_lgbm_model, additional_iterations, String[], verbosity, LightGBM.Dates.now())
         # should probably dump report too, its just an empty dict...
         fitresult = (old_lgbm_model, old_classes, mlj_model)
-        return (fitresult, cache, report)
+        return (fitresult, cache, (report,))
     end
```
ablaom, line 189: If there is no report, you can just return `NamedTuple()`. Otherwise it should be a named-tuple with informative keys.

src/MLJInterface.jl
```diff
@@ -183,17 +183,35 @@ function update(mlj_model::MLJInterface.MODELS, verbosity::Int, fitresult, cache
...
+        # if we get here it's just means we need to call fit directly
+        return MLJInterface.fit(mlj_model, verbosity, X, y, w)
     end
 
-    # if we get here it's just means we need to call fit directly
-    return MLJInterface.fit(mlj_model, verbosity, X, y, w)
+    additional_iterations = mlj_model.num_iterations - old_mlj_model.num_iterations
+    if verbosity >= 1
+        @info("Not refitting from scratch", additional_iterations)
+    end
```
ablaom, line 194: Something missing in the log here? Like "Not refitting from scratch. Adding $additional_iterations iterations. "

src/MLJInterface.jl
```diff
@@ -157,7 +157,7 @@ function mlj_to_kwargs(model::LGBMClassifier, classes)
 end
 
 # X and y and w must be untyped per MLJ docs
-function fit(mlj_model::MODELS, verbosity::Int, X, y, w=Vector{AbstractFloat}())
+function fit(mlj_model::MODELS, verbosity::Int, X, y, w=AbstractFloat[])
```
iqml, line 160: Just a style change? If so, you have others.

src/fit.jl
```diff
@@ -85,9 +97,7 @@ function train!(estimator::LGBMEstimator, tests_names::Vector{String}, verbosity
                      "split requirements.")
         end
         if is_finished == 1
-            # save the model in serialised form, in case we should be deepcopied or serialised elsewhere
-            estimator.model = LGBM_BoosterSaveModelToString(estimator.booster, 0, 0)
```
iqml, line 89: Please add commentary on why this was removed.

test/runtests.jl
```diff
@@ -32,6 +32,10 @@ end
         include(joinpath("mlj", "regression.jl"))
     end
 
+    @testset "MLJ update interface" begin
+        include(joinpath("mlj", "update.jl"))
```
iqml, line 36: You have a `true` at the end of the module but you're not calling `@test` on the `include`?

src/MLJInterface.jl
iqml: `fit.jl` 30-31

src/fit.jl
yalwan-iqvia: Its the same as what happens a few lines below when we exit the function. When I first added the savemodel to string this bit me, and i copied it up there, and while developing this now I realised its much better to break form the loop and have all the end-loop logic at one place only (saves duplication too)

src/MLJInterface.jl
yalwan-iqvia: Log lines line this will actually do something of the form:
```
julia> @info "hello" buffer
┌ Info: hello
└   buffer = "buffer"

julia> 

```

test/runtests.jl
yalwan-iqvia: Yes I think I discovered it wasn't necessary and forgot to dump the true's

src/MLJInterface.jl
yalwan-iqvia: Thanks -- something about `ArrayType{Type}(initialise)` grinds my gears and i find `TypeToArray[]` much cleaner

src/MLJInterface.jl
yalwan-iqvia: So I think there is in theory a report if user specifies the metrics that underlying library evaluates but I never fully checked it, and I'd rather not mess around with it now

src/MLJInterface.jl
yalwan-iqvia: What this is doing is catching the instance where MLJ defaulting has provided nothing, and on right hand side you can see it just calls function without weights parameter

Line below is where we catch what MLJ passes to us (that isn't `nothing` and we pass it through. I've added tests though.

yalwan-iqvia: 25ee0d4

src/MLJInterface.jl
yalwan-iqvia: 25ee0d4

src/MLJInterface.jl
yalwan-iqvia: 25ee0d4

src/MLJInterface.jl
ablaom: Great to know, thanks!

src/MLJInterface.jl
ablaom: Yes, you're right. I see there is a line above which covers the other case.

Issue #45, opened 2020-05-01
Implement LGBM_BoosterFeatureImportance and expose feature importance


aviatesk:
how about `feature_importance(model)` functionality to provide information about features that were "important" while fitting ?

Maybe [these lines](https://github.com/microsoft/LightGBM/blob/7076cb8a3ac3a7b32dcf37be5593dddf27bf7f16/python-package/lightgbm/basic.py#L2759-L2795) will help to implement this.

yalwan-iqvia:
Thanks, this looks like a missing FFI call. Do you mind if I change the title of the issue?

aviatesk:
> Do you mind if I change the title of the issue?

never ;)

yalwan-iqvia:
So it is ok? 😁

aviatesk:
ah sorry for my bad engligh, no, I don't mind if you change the title, of course 😅

yalwan-iqvia:
Your English is fine, I just wanted to be certain 👍

yalwan-iqvia:
@aviatesk and anybody else who checks this issue

As a workaround, for now, one can do `mean(LightGBM.predict(e, X; predict_type=3); dims=1)`, which will average SHAP values for all preidctions -- The C-API supports it even though our documentation does not mention this.

So I will also update documentation for `predict` when I develop this feature.

yalwan-iqvia:
PR for the feature, sans MLJ integration: https://github.com/IQVIA-ML/LightGBM.jl/pull/52

Does this look sensible to you?

yalwan-iqvia:
@aviatesk 

I've merged said PR https://github.com/IQVIA-ML/LightGBM.jl/pull/52

It implements `gain_importance` and `split_importance` functions for the medium level API, and `LGBM_BoosterFeatureImportance` from the low level (FFI) API. There is no MLJ integration yet, if you have any suggestions it would be appreciated.

I'd like to close this issue and have new one for the MLJ integration aspect, if you think that is ok.

aviatesk:
Ah, thanks for addressing this and letting me know the update !
I will try to take a look at your PR tonight, but I'm not sure I can post meaningful feedback on this.
Well, I'm totally okay if you close this issue :)

yalwan-iqvia:
Raised #61

PR #52, opened 2020-06-01
adding feature importance function

yalwan-iqvia:
Part of: https://github.com/IQVIA-ML/LightGBM.jl/issues/45

MLJ integration remains, for another PR


src/utils.jl
```diff
@@ -12,6 +12,24 @@ const INDEXPARAMS = [:categorical_feature]
 
 const MAXIMIZE_METRICS = ["auc", "ndcg"]
 
+# LOGGING Funcs
```
yalwan-iqvia, line 15: This is just rearranging

PR #62, opened 2020-06-24
Add dart goss

sbeura:
- adding support for DART and GOSS boosting.


README.md
```diff
@@ -274,6 +274,14 @@ LGBMRegression(;
     save_binary = false,
     categorical_feature = Int[],
     is_unbalance = false,
+    drop_rate = 0.1,
```
yalwan-iqvia, line 277: Missed addding `boosting` parameter in README

README.md
```diff
@@ -316,6 +324,14 @@ LGBMClassification(;
...
```
yalwan-iqvia, line 327: Same comment as above

README.md
```diff
@@ -274,6 +275,14 @@ LGBMRegression(;
...
+    max_drop = 50,
+    skip_drop = 0.5,
+    xgboost_dart_mode = false,
```
kd-iqvia, line 281: I saw it on the LightGBM params that this flag defaults to false and it's only used when `boosting = "dart"`. It's just out of curiosity - is there a way to get rid of this flag and just have whatever is indicated in `boosting`?

yalwan-iqvia: >is there a way to get rid of this flag and just have whatever is indicated in boosting?

Do you mean by this to have it implied by `boosting = "dart"`? If so, no, because just because DART is on it doesn't necessarily mean you want the XGBoost mode. Additionally it would require a lot of logic of the kind we are trying to avoid here, especially as a lot of it would essentially be duplicating logic already found in C-API, so in general we try to avoid parameter logic and just let API throw errors if necessary

kd-iqvia: yes I thought about having it implied by `boosting = "dart"` but yes I didn't see the whole complexity around it (and duplication). it's fine as it is.

PR #66, opened 2020-09-15
MLJ: Feature importances in user fitreport

yalwan-iqvia:
This is the first real use of the user report containing anything of note

* Integrated feature importances into  the report
* Added tests to try and control the shape of the data structure (namedtuple) and ensure breaking changes don't occur in future (this is a breaking change as it is)
* Implemented a metrics merger function as a result of supporting `MLJ.update` without losing previous metrics history
* Reworked how metrics storage is computed in light of metrics being evaluated over partial runs (some iterations here, some more iterations there)

### DIFF
--- a/src/MLJInterface.jl
+++ b/src/MLJInterface.jl
@@ -153,6 +153,7 @@ function mlj_to_kwargs(model::LGBMRegressor)
 
 end
 
+
 function mlj_to_kwargs(model::LGBMClassifier, classes)
 
     num_class = length(classes)
@@ -174,6 +175,7 @@ function mlj_to_kwargs(model::LGBMClassifier, classes)
 
 end
 
+
 # X and y and w must be untyped per MLJ docs
 function fit(mlj_model::MODELS, verbosity::Int, X, y, w=AbstractFloat[])
 
@@ -186,15 +188,18 @@ function fit(mlj_model::MODELS, verbosity::Int, X, y, w=AbstractFloat[])
     X = MLJModelInterface.matrix(X)
     # The FFI wrapper wants Float32 for these
     w = Float32.(w)
-    report = LightGBM.fit!(model, X, y_lgbm; verbosity=verbosity, weights=w)
+    train_metrics = LightGBM.fit!(model, X, y_lgbm; verbosity=verbosity, weights=w)
 
     fitresult = (model, classes, deepcopy(mlj_model))
-    cache = (num_boostings_done=[LightGBM.get_iter_number(model)], )
-    report = (report,)
+    # because update needs access to the older version of training metrics we keep them in the cache
+    # so the update can merge old and additional metrics as necessary.
+    cache = (num_boostings_done=[LightGBM.get_iter_number(model)], training_metrics=train_metrics)
+    report = user_fitreport(model, train_metrics)
 
     return (fitresult, cache, report)
 
 end
+
 
 function update(mlj_model::MLJInterface.MODELS, verbosity::Int, fitresult, cache, X, y, w=AbstractFloat[])
 
@@ -228,17 +233,16 @@ function update(mlj_model::MLJInterface.MODELS, verbosity::Int, fitresult, cache
     old_lgbm_model.num_iterations = num_iterations + additional_iterations
 
     # eh this is ugly, possibly prompts a need for some refactoring
-    report = LightGBM.train!(old_lgbm_model, additional_iterations, String[], verbosity, LightGBM.Dates.now())
-
-    # should probably dump report too, its just an empty dict...
+    train_metrics = LightGBM.train!(old_lgbm_model, additional_iterations, String[], verbosity, LightGBM.Dates.now())
     fitresult = (old_lgbm_model, old_classes, deepcopy(mlj_model))
 
     final_num_iter = LightGBM.get_iter_number(old_lgbm_model)
     iteration_history = deepcopy(cache.num_boostings_done)
     push!(iteration_history, final_num_iter - num_iterations)
-    newcache = (num_boostings_done=iteration_history, )
+    report = user_fitreport(old_lgbm_model, cache.training_metrics, train_metrics)
+    newcache = (num_boostings_done=iteration_history, training_metrics=report.training_metrics)
 
-    return (fitresult, newcache, (report,))
+    return (fitresult, newcache, report)
 
 end
 
@@ -270,10 +274,27 @@ function predict_classifier((fitted_model, classes, _), Xnew)
 
 end
 
+
 function predict_regression((fitted_model, classes, _), Xnew)
 
     Xnew = MLJModelInterface.matrix(Xnew)
     return dropdims(LightGBM.predict(fitted_model, Xnew), dims=2)
+
+end
+
+# This function returns a user accessible report and therefore needs to not be a source of breaking changes
+# Keep all of the functinality required to generate this into a single function so that it can be
+# kept under control via tests and so on. It's not a lot for now but is likely to grow.
+function user_fitreport(estimator::LightGBM.LGBMEstimator, eval_metrics::Dict)
+
+    importance = (gain = LightGBM.gain_importance(estimator), split = LightGBM.split_importance(estimator))
+    return (training_metrics = deepcopy(eval_metrics), importance = importance)
+
+end
+function user_fitreport(estimator::LightGBM.LGBMEstimator, eval_metrics::Dict, new_eval_metrics::Dict)
+
+    metrics = LightGBM.merge_scores(eval_metrics, new_eval_metrics)
+    return user_fitreport(estimator, metrics)
 
 end
 
--- a/src/fit.jl
+++ b/src/fit.jl
@@ -135,10 +135,10 @@ function eval_metrics!(results::Dict{String,Dict{String,Vector{Float64}}},
         end
     end
 
+    # Metrics for test sets
     if (iter - 1) % estimator.metric_freq == 0 || estimator.early_stopping_round > 0
         for (test_idx, test_name) in enumerate(tests_names)
             scores = LGBM_BoosterGetEval(estimator.booster, test_idx)
-
             # Check if progress should be stored and/or printed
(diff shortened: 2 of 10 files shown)


src/MLJInterface.jl
```diff
@@ -186,16 +188,19 @@ function fit(mlj_model::MODELS, verbosity::Int, X, y, w=AbstractFloat[])
     X = MLJModelInterface.matrix(X)
     # The FFI wrapper wants Float32 for these
     w = Float32.(w)
-    report = LightGBM.fit!(model, X, y_lgbm; verbosity=verbosity, weights=w)
+    train_metrics = LightGBM.fit!(model, X, y_lgbm; verbosity=verbosity, weights=w)
```
yalwan-iqvia, line 191: This is so called because via MLJ it isn't possible (as designed so far) to get validation metrics (after all, one would use MLJ for the functionality)

src/fit.jl
```diff
@@ -91,16 +91,23 @@ function train!(
 
     start_iter = get_iter_number(estimator) + 1
     end_iter = start_iter + num_iterations - 1
+    total_metrics_evals = sum((((start_iter:end_iter) .- 1) .% estimator.metric_freq) .== 0)
```
yalwan-iqvia, line 94: We need to compute this (and some other things down the line) because of the way store scores works, it was getting OOB error with previous implemention because it assumed all trainings start from scratch which wasn't true since `update` was implemented for MLJ. This one tells us the size of the array to allocate in store scores

src/fit.jl
```diff
@@ -91,16 +91,23 @@ function train!(
...
 
     for iter in start_iter:end_iter
+
         is_finished = LGBM_BoosterUpdateOneIter(estimator.booster)
-        log_debug(verbosity, Dates.CompoundPeriod(now() - start_time),
-                  " elapsed, finished iteration ", iter, "\n")
+
+        log_debug(verbosity, Dates.CompoundPeriod(now() - start_time), " elapsed, finished iteration ", iter, "\n")
+
+        metric_idx = sum((((start_iter:iter) .- 1) .% estimator.metric_freq) .== 0)
```
yalwan-iqvia, line 102: This one tells us the index to save into for store scores. We're only saving every metric_freq iterations and we're only saving for the batch of iterations we're currently doing

src/fit.jl
```diff
@@ -91,16 +91,23 @@ function train!(
...
+
         if is_finished == 0
-            is_finished = eval_metrics!(results, estimator, tests_names, iter, n_metrics,
-                                        verbosity, bigger_is_better, best_score, best_iter, metrics)
+            is_finished = eval_metrics!(
```
yalwan-iqvia, line 105: reformatting and also passing out new size & index values into function

src/fit.jl
```diff
@@ -91,16 +91,23 @@ function train!(
...
+                results, estimator, tests_names, iter, metric_idx, total_metrics_evals, n_metrics,
+                verbosity, bigger_is_better, best_score, best_iter, metrics,
+            )
         else
-            shrinkresults!(results, iter - 1)
+            shrinkresults!(results, metric_idx)
```
yalwan-iqvia, line 110: Same reasoning applies to results shrinkage

src/fit.jl
```diff
@@ -122,26 +129,36 @@ function train!(
 end
 
 
-function eval_metrics!(results::Dict{String,Dict{String,Vector{Float64}}},
-                       estimator::LGBMEstimator, tests_names::Vector{String}, iter::Integer,
-                       n_metrics::Integer, verbosity::Integer,
-                       bigger_is_better::Vector{Float64}, best_score::Matrix{Float64},
-                       best_iter::Matrix{Int}, metrics::Vector{String})
+function eval_metrics!(
```
yalwan-iqvia, line 132: reformatting, adding new args metrics_store_idx and metrics_stored_count

src/fit.jl
```diff
@@ -122,26 +129,36 @@ function train!(
...
+    best_score::Matrix{Float64},
+    best_iter::Matrix{Int},
+    metrics::Vector{String},
+)
+
     if (iter - 1) % estimator.metric_freq == 0
         if estimator.is_training_metric
             scores = LGBM_BoosterGetEval(estimator.booster, 0)
-            store_scores!(results, estimator, iter, "training", scores, metrics)
+            store_scores!(results, estimator, metrics_store_idx, metrics_stored_count, "training", scores, metrics)
```
yalwan-iqvia, line 150: passing new args through

src/fit.jl
```diff
@@ -122,26 +129,36 @@ function train!(
...
 
+    # Metrics for test sets
     if (iter - 1) % estimator.metric_freq == 0 || estimator.early_stopping_round > 0
         for (test_idx, test_name) in enumerate(tests_names)
             scores = LGBM_BoosterGetEval(estimator.booster, test_idx)
-
             # Check if progress should be stored and/or printed
             if (iter - 1) % estimator.metric_freq == 0
-                store_scores!(results, estimator, iter, test_name, scores, metrics)
+                store_scores!(results, estimator, metrics_store_idx, metrics_stored_count, test_name, scores, metrics)
```
yalwan-iqvia, line 161: passing new args through

src/fit.jl
```diff
@@ -152,7 +169,9 @@ function eval_metrics!(results::Dict{String,Dict{String,Vector{Float64}}},
                     best_score[metric_idx, test_idx] = maximize_score
                     best_iter[metric_idx, test_idx] = iter
                 elseif iter - best_iter[metric_idx, test_idx] >= estimator.early_stopping_round
-                    shrinkresults!(results, best_iter[metric_idx, test_idx])
+                    # need to fix this, its clearly broken in light of latest updates
+                    # This will shrink it up to the current stored metric
+                    shrinkresults!(results, metrics_store_idx)
```
yalwan-iqvia, line 174: I can't exactly put my finger on why but this is incorrect (in the sense that the previous logic was wanting to shrink values to best iteration and here we just shrink if we're about to stop to current storage index)

src/fit.jl
```diff
@@ -165,26 +184,65 @@ function eval_metrics!(results::Dict{String,Dict{String,Vector{Float64}}},
 end
 
 
-function store_scores!(results::Dict{String,Dict{String,Vector{Float64}}},
-                       estimator::LGBMEstimator, iter::Integer, evalname::String,
-                       scores::Vector{Cdouble}, metrics::Vector{String})
+function store_scores!(
```
yalwan-iqvia, line 187: reformatted, and eating the new args (plus num_evals didnt need computing inside function as its being passed as an arg now)

src/fit.jl
```diff
@@ -165,26 +184,65 @@ function eval_metrics!(results::Dict{String,Dict{String,Vector{Float64}}},
...
-        eval_idx = cld(iter, estimator.metric_freq)
-        results[evalname][metric_name][eval_idx] = scores[metric_idx]
+        results[evalname][metric_name][store_idx] = scores[metric_idx]
     end
 
     return nothing
 end
 
 
+function merge_scores(
```
yalwan-iqvia, line 211: This is the functionality for merging metrics histories if training is completed in multiple batches

src/fit.jl
yalwan-iqvia: I think this is ok -- it might have always been wrong with metrics_freq > 1

src/fit.jl
```diff
@@ -92,15 +92,25 @@ function train!(
     start_iter = get_iter_number(estimator) + 1
     end_iter = start_iter + num_iterations - 1
 
-    for iter in start_iter:end_iter
+    metrics_idx_sequence = (((start_iter:end_iter) .- 1) .% estimator.metric_freq) .== 0
```
iqml, line 95: You can do arithmetic on ranges without collection, i.e. you don't need to broadcast here until the equality.
```julia
metrics_idx_sequence = ((start_iter:end_iter - 1) % estimator.metric_freq) .== 0
```

src/fit.jl
```diff
@@ -92,15 +92,25 @@ function train!(
...
+
+    for (idx, iter) in enumerate(start_iter:end_iter)
+
         is_finished = LGBM_BoosterUpdateOneIter(estimator.booster)
-        log_debug(verbosity, Dates.CompoundPeriod(now() - start_time),
-                  " elapsed, finished iteration ", iter, "\n")
+
+        log_debug(verbosity, Dates.CompoundPeriod(now() - start_time), " elapsed, finished iteration ", iter, "\n")
+
+        metric_idx = metric_idx + metrics_idx_sequence[idx]
```
iqml, line 105: `+=` ?

src/fit.jl
yalwan-iqvia: The example code you provided it subtracts 1 from end_iter as opposed to the whole range
When the range is put into parenthesis then nonbroadcast subtraction fails with no such method

```
julia> (1:5) -1
ERROR: MethodError: no method matching -(::UnitRange{Int64}, ::Int64)
Closest candidates are:
  -(::Complex{Bool}, ::Real) at complex.jl:307
  -(::Missing, ::Number) at missing.jl:115
  -(::Base.CoreLogging.LogLevel, ::Integer) at logging.jl:108
  ...
Stacktrace:
 [1] top-level scope at REPL[1]:1

julia> (1:5 -1)
1:4

julia> (1:5 -1) % 2
ERROR: MethodError: no method matching rem(::UnitRange{Int64}, ::Int64)
Closest candidates are:
  rem(::Any, ::Any, ::RoundingMode{:ToZero}) at div.jl:67
  rem(::Any, ::Any, ::RoundingMode{:Down}) at div.jl:68
  rem(::Any, ::Any, ::RoundingMode{:Up}) at div.jl:69
  ...
Stacktrace:
 [1] top-level scope at REPL[3]:1

julia> (1:5 -1) .% 2
4-element Array{Int64,1}:
 1
 0
 1
 0

julia> 
```

src/fit.jl
yalwan-iqvia: I don't personally prefer this way of doing it, so I'll leave it In this case I think

PR #68, opened 2020-09-24
v.3.0.0 compatibility

### LINKED ISSUE #65: LightGBM 3.x.x support
LightGBM recently released 3.0.0 version. Make sure wrapper supports this version. Bonus points if we are able to keep supporting both 2.3.1 and 3.0.0 simultaneously.

kd-iqvia:
resolves #65 
sadly can't keep both v2.3.1 and v3.0.0 due to implementation of  few breaking changes that had to be ported here so no bonus point for me 😜

### DIFF
--- a/src/MLJInterface.jl
+++ b/src/MLJInterface.jl
@@ -6,7 +6,7 @@ import LightGBM
 
 const LGBM_METRICS = (
     "None", "l1", "l2", "rmse", "quantile", "mape", "huber", "fair", "poisson", "gamma", "gamma_deviance",
-    "tweedie", "ndcg", "lambdarank", "map", "mean_average_precision", "auc", "binary_logloss", "binary",
+    "tweedie", "ndcg", "lambdarank", "map", "mean_average_precision", "auc", "average_precision", "binary_logloss", "binary",
     "binary_error", "auc_mu", "multi_logloss", "multi_error", "cross_entropy", "xentropy", "multi_logloss",
     "multiclass", "softmax", "multiclassova", "multiclass_ova", "ova", "ovr", "cross_entropy_lambda",
     "xentlambda", "kullback_leibler", "kldiv",
@@ -47,7 +47,7 @@ MLJModelInterface.@mlj_model mutable struct LGBMRegressor <: MLJModelInterface.D
     max_bin::Int = 255::(_ > 1)
     init_score::String = ""
     drop_rate::Float64 = 0.1 :: (0.0 <= _ <= 1.0)
-    max_drop::Int = 50 
+    max_drop::Int = 50
     skip_drop:: Float64 = 0.5 :: (0.0 <= _ <= 1)
     xgboost_dart_mode::Bool
     uniform_drop::Bool
@@ -104,7 +104,7 @@ MLJModelInterface.@mlj_model mutable struct LGBMClassifier <: MLJModelInterface.
     max_bin::Int = 255::(_ > 1)
     init_score::String = ""
     drop_rate::Float64 = 0.1 :: (0.0 <= _ <= 1.0)
-    max_drop::Int = 50 
+    max_drop::Int = 50
     skip_drop:: Float64 = 0.5 :: (0.0 <= _ <= 1)
     xgboost_dart_mode::Bool
     uniform_drop::Bool
--- a/src/MLJInterface.jl
+++ b/src/MLJInterface.jl
@@ -6,14 +6,13 @@ import LightGBM
 
 const LGBM_METRICS = (
     "None", "l1", "l2", "rmse", "quantile", "mape", "huber", "fair", "poisson", "gamma", "gamma_deviance",
-    "tweedie", "ndcg", "lambdarank", "map", "mean_average_precision", "auc", "average_precision", "binary_logloss", "binary",
-    "binary_error", "auc_mu", "multi_logloss", "multi_error", "cross_entropy", "xentropy", "multi_logloss",
-    "multiclass", "softmax", "multiclassova", "multiclass_ova", "ova", "ovr", "cross_entropy_lambda",
+    "tweedie", "ndcg", "lambdarank", "map", "mean_average_precision", "auc", "average_precision", "binary_logloss",
+    "binary_error", "auc_mu", "multi_logloss", "multi_error", "cross_entropy", "xentropy", "multi_logloss","cross_entropy_lambda",
     "xentlambda", "kullback_leibler", "kldiv",
 )
 
 const CLASSIFICATION_OBJECTIVES = (
-    "binary", "multiclass", "softmax", "multiclassova", "multiclass_ova", "ova", "ovr",
+    "binary", "multiclass", "softmax",
 )
 
 const REGRESSION_OBJECTIVES = (
--- a/src/fit.jl
+++ b/src/fit.jl
@@ -120,7 +120,7 @@ function train!(
     end
 
     # save the model in serialised form, in case we should be deepcopied or serialised elsewhere
-    estimator.model = LGBM_BoosterSaveModelToString(estimator.booster, 0, 0)
+    estimator.model = LGBM_BoosterSaveModelToString(estimator.booster, 0, 0, 0)
 
     return results
 end
--- a/src/fit.jl
+++ b/src/fit.jl
@@ -266,7 +266,7 @@ function stringifyparams(estimator::LGBMEstimator, params::Vector{Symbol})
             if !isempty(param_value)
                 # Convert parameters that contain indices to C's zero-based indices.
                 if in(param_name, INDEXPARAMS)
-                    param_value -= 1
+                    param_value .-= 1
                 end
 
                 if typeof(param_value) <: Array
--- a/src/fit.jl
+++ b/src/fit.jl
@@ -120,7 +120,7 @@ function train!(
     end
 
     # save the model in serialised form, in case we should be deepcopied or serialised elsewhere
-    estimator.model = LGBM_BoosterSaveModelToString(estimator.booster, 0, 0, 0)
+    estimator.model = LGBM_BoosterSaveModelToString(estimator.booster)
 
     return results
 end
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -12,7 +12,7 @@ const BOOSTERPARAMS = [
 
 const INDEXPARAMS = [:categorical_feature]
 
-const MAXIMIZE_METRICS = ["auc", "ndcg"]
+const MAXIMIZE_METRICS = ["auc", "ndcg", "average_precision"]
 
 # LOGGING Funcs
 function log_fatal(verbosity::Integer, msg...)
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -42,8 +42,17 @@ Save the fitted model in `estimator` as `filename`.
 * `filename::String`: the name of the file to save the model in.
 * `num_iteration::Integer`: keyword argument that sets the number of iterations of the model that
     should be saved. `< 0` for all iterations.
+* `start_iteration` : : Start index of the iteration that should be saved.
+* `feature_importance_type` : Type of feature importance,
+    can be C_API_FEATURE_IMPORTANCE_SPLIT or C_API_FEATURE_IMPORTANCE_GAIN
 """
-function savemodel(estimator::LGBMEstimator, filename::String; num_iteration::Integer=-1, start_iteration::Integer=0, feature_importance_type::Integer=0)
+function savemodel(
+    estimator::LGBMEstimator,
+    filename::String;
+    num_iteration::Integer=-1,
+    start_iteration::Integer=0,
+    feature_importance_type::Integer=0
+)
     @assert(estimator.booster.handle != C_NULL, "Estimator does not contain a fitted model.")
     LGBM_BoosterSaveModel(estimator.booster, start_iteration, num_iteration, feature_importance_type, filename)
     return nothing
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -43,9 +43,9 @@ Save the fitted model in `estimator` as `filename`.
 * `num_iteration::Integer`: keyword argument that sets the number of iterations of the model that
     should be saved. `< 0` for all iterations.
 """
-function savemodel(estimator::LGBMEstimator, filename::String; num_iteration::Integer = -1,start_iteration::Integer=0)
+function savemodel(estimator::LGBMEstimator, filename::String; num_iteration::Integer = -1, start_iteration::Integer=0, feature_importance_type::Integer=0)
     @assert(estimator.booster.handle != C_NULL, "Estimator does not contain a fitted model.")
-    LGBM_BoosterSaveModel(estimator.booster, start_iteration , num_iteration , filename)
+    LGBM_BoosterSaveModel(estimator.booster, start_iteration, num_iteration, feature_importance_type, filename)
     return nothing
 end
 
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -43,7 +43,7 @@ Save the fitted model in `estimator` as `filename`.
 * `num_iteration::Integer`: keyword argument that sets the number of iterations of the model that
     should be saved. `< 0` for all iterations.
 """
-function savemodel(estimator::LGBMEstimator, filename::String; num_iteration::Integer = -1, start_iteration::Integer=0, feature_importance_type::Integer=0)
+function savemodel(estimator::LGBMEstimator, filename::String; num_iteration::Integer=-1, start_iteration::Integer=0, feature_importance_type::Integer=0)
     @assert(estimator.booster.handle != C_NULL, "Estimator does not contain a fitted model.")
     LGBM_BoosterSaveModel(estimator.booster, start_iteration, num_iteration, feature_importance_type, filename)
     return nothing
--- a/src/wrapper.jl
+++ b/src/wrapper.jl
@@ -217,7 +217,7 @@ function LGBM_DatasetGetFeatureNames(ds::Dataset)
                 new_buffer_len => Csize_t,
                 out_buffer_len => Ref{Csize_t},
                 feature_names => Ref{Ptr{UInt8}})
-
+  
     return [unsafe_string(pointer(feature_name)) for feature_name in feature_names[1:num_feature_names[]]]
 end
 
--- a/src/wrapper.jl
+++ b/src/wrapper.jl
@@ -60,7 +60,7 @@ function Base.deepcopy_internal(x::Booster, stackdict::IdDict)
         return Booster()
     end
 
-    serialised = LGBM_BoosterSaveModelToString(x, 0, 0)
+    serialised = LGBM_BoosterSaveModelToString(x, 0, 0, 0)
     y = LGBM_BoosterLoadModelFromString(serialised)
 
     return y
@@ -135,7 +135,7 @@ end
 
 function LGBM_DatasetCreateFromMat(data::Matrix{T}, parameters::String,
                                                               reference::Dataset,
-                                                              is_row_major::Bool = false) where T<:Union{Float32,Float64} 
+                                                              is_row_major::Bool = false) where T<:Union{Float32,Float64}
     lgbm_data_type = jltype_to_lgbmid(T)
     nrow, ncol = ifelse(is_row_major, reverse(size(data)), size(data))
     out = Ref{DatasetHandle}()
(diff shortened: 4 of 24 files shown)


test/ffi/datasets.jl
```diff
@@ -108,15 +108,16 @@ end
...
-#     # test string overlengths
-#     name_1 = randstring(3000)
-#     name_2 = randstring(3000)
-#
-#     @test LightGBM.LGBM_DatasetSetFeatureNames(created_dataset, [name_1, name_2]) == nothing
-#     retrieved = LightGBM.LGBM_DatasetGetFeatureNames(created_dataset)
-#
-#     @test retrieved[1] == name_1#[1:256]
-#     @test retrieved[2] == name_2#[1:256]
+    # test string overlengths
```
kd-iqvia, line 111: used to give segfault before now works

test/runtests.jl
```diff
@@ -24,53 +24,53 @@ end
         include(joinpath("mlj", "parameters.jl"))
     end
 
-    @testset "Binary LightGBM" begin
```
kd-iqvia, line 27: still working on all the commented ones

test/ffi/booster.jl
```diff
@@ -295,19 +295,17 @@ end
...
 
     booster = LightGBM.LGBM_BoosterCreateFromModelfile(joinpath(@__DIR__, "data", "test_tree"))
-    string_repr = LightGBM.LGBM_BoosterSaveModelToString(booster, 0, 0)
+    string_repr = LightGBM.LGBM_BoosterSaveModelToString(booster, 0, 0, 0)
     # so it turns out that the string save and file save aren't necesarily the same so..
     # check a bunch of expected substrings, etc
     @test occursin("version=v3", string_repr)
     @test occursin("num_leaves=1", string_repr)
     @test occursin("end of trees", string_repr)
-    @test occursin("feature importances:", string_repr)
```
kd-iqvia, line 304: `feature importances` are no longer present in the string save

src/fit.jl
```diff
@@ -120,7 +120,7 @@ function train!(
     end
 
     # save the model in serialised form, in case we should be deepcopied or serialised elsewhere
-    estimator.model = LGBM_BoosterSaveModelToString(estimator.booster, 0, 0)
+    estimator.model = LGBM_BoosterSaveModelToString(estimator.booster, 0, 0, 0)
```
kd-iqvia, line 123: need to re-think usage of the new parameter. as per API update it's the `feature_importance_type` parameter indicating the type of feature importance, where `0` is for `C_API_FEATURE_IMPORTANCE_SPLIT`and `1` for `C_API_FEATURE_IMPORTANCE_GAIN`

src/wrapper.jl
```diff
@@ -189,13 +189,35 @@ function LGBM_DatasetSetFeatureNames(ds::Dataset, feature_names::Vector{String})
...
-    n_features = LGBM_DatasetGetNumFeature(ds)
-    feature_names = [Vector{UInt8}(undef,256) for i in 1:n_features]
+    len = Cint(2)
+    buffer_len = Csize_t(2)
+    feature_names = [Vector{UInt8}(undef, buffer_len) for i in 1:len]
+
+   # setting these so that the first C API call informs us of their allocations
     num_feature_names = Ref{Cint}()
+    out_buffer_len = Ref{Csize_t}()
+
```
kd-iqvia, line 199: oopsie additional empty line not needed

src/wrapper.jl
```diff
@@ -468,12 +534,16 @@ end
 function LGBM_BoosterCalcNumPredict(bst::Booster, num_row::Integer, predict_type::Integer,
                                     num_iteration::Int)
     out_len = Ref{Int64}()
+    start_iteration = Cint(1)
```
kd-iqvia, line 537: should have added it as a param instead and use as 0 like in `LGBM_BoosterSaveModel`

src/wrapper.jl
```diff
@@ -398,25 +420,69 @@ function LGBM_BoosterGetEvalCounts(bst::Booster)
 end
 
 function LGBM_BoosterGetEvalNames(bst::Booster)
```
kd-iqvia, line 422: need to enable test for this one and the `LGBM_BoosterGetFeatureNames`

src/wrapper.jl
```diff
@@ -398,25 +420,69 @@ function LGBM_BoosterGetEvalCounts(bst::Booster)
...
+              buffer_len => Csize_t,
+              out_buffer_len => Ref{Csize_t},
               out_strs => Ref{Ptr{UInt8}})
+
+    # allocating memory
+    new_len = out_len[]
+    new_buffer_len = out_buffer_len[]
+    out_strs = [Vector{UInt8}(undef, new_buffer_len) for i in 1:new_len]
+
+    @lightgbm(:LGBM_BoosterGetEvalNames,
```
kd-iqvia, line 478: oopsie :)

test/ffi/booster.jl
kd-iqvia: brought it back now with the `_`

test/ffi/datasets.jl
```diff
@@ -108,15 +108,16 @@ end
     @test LightGBM.LGBM_DatasetGetFeatureNames(created_dataset) == fieldnames
 
     # This stuff makes it segfault so leaving it alone...
```
kd-iqvia, line 110: need to delete this comment no longer relevant

test/mlj/binary_classifier.jl
```diff
@@ -38,7 +38,7 @@ misclassification_rate   = sum(yhat .!= y[test])/length(test)
 @test misclassification_rate < 0.015
 
 # All we can really say about fitting with/without weights for this example is that the solutions shouldn't be identical
-@test yhat_with_weights != yhat
+# @test yhat_with_weights != yhat
```
kd-iqvia, line 41: i'm not leaving it like that 😸

kd-iqvia: the drama here is that the fitting with and without weights shouldn't be identical... but... they are somehow identical...

test/mlj/binary_classifier.jl
```diff
@@ -30,15 +30,14 @@ yhat                     = MLJBase.mode.(MLJBase.predict(model, fitresult, MLJBa
...
-
 misclassification_rate   = sum(yhat .!= y[test])/length(test)
 
 # Well, although XGBoost gets misclassification below 0.01, LightGBM can't do it with the default settings ...
 # It gets to exactly 0.01...
 @test misclassification_rate < 0.015
 
 # All we can really say about fitting with/without weights for this example is that the solutions shouldn't be identical
-@test yhat_with_weights != yhat
+@test_broken yhat_with_weights != yhat
```
kd-iqvia, line 40: Currently, the status has been changed to `@test_broken` .  After implementing the API changes, the outputs of fitting with and without weights seem to be the same which is not the expected behaviour. This problem only occurs in case of `binary classification` and is not present for `multiclass`, or `regression` (where I have added a similar test).

src/fit.jl
kd-iqvia: I added a test to cover this case and it seems there is no difference if either of them is provided as a parameter. The test generates a model, calculates split and gain, then uses the `LGBM_BoosterSaveModelToString` to same model and then to re-loads it. At each stage the split and gain are checked vs the expected outputs.

src/utils.jl
```diff
@@ -43,9 +43,9 @@ Save the fitted model in `estimator` as `filename`.
 * `num_iteration::Integer`: keyword argument that sets the number of iterations of the model that
     should be saved. `< 0` for all iterations.
 """
-function savemodel(estimator::LGBMEstimator, filename::String; num_iteration::Integer = -1,start_iteration::Integer=0)
+function savemodel(estimator::LGBMEstimator, filename::String; num_iteration::Integer=-1, start_iteration::Integer=0, feature_importance_type::Integer=0)
```
yalwan-iqvia, line 46: Should probably consider reformatting the line like this:

```
function savemodel(
    estimator::LGBMEstimator,
    filename::String; num_iteration::Integer=-1,
    start_iteration::Integer=0,
    feature_importance_type::Integer=0
)
```

src/wrapper.jl
```diff
@@ -135,7 +135,7 @@ end
 
 function LGBM_DatasetCreateFromMat(data::Matrix{T}, parameters::String,
```
yalwan-iqvia, line 136: Consider reformatting this line

src/wrapper.jl
```diff
@@ -189,13 +189,34 @@ function LGBM_DatasetSetFeatureNames(ds::Dataset, feature_names::Vector{String})
...
-    feature_names = [Vector{UInt8}(undef,256) for i in 1:n_features]
+    len = Cint(2)
+    buffer_len = Csize_t(2)
+    feature_names = [Vector{UInt8}(undef, buffer_len) for i in 1:len]
+
+   # setting these so that the first C API call informs us of their allocations
     num_feature_names = Ref{Cint}()
+    out_buffer_len = Ref{Csize_t}()
+
     @lightgbm(:LGBM_DatasetGetFeatureNames,
```
yalwan-iqvia, line 200: Consider the alternative formatting:

```
    @lightgbm(
        :LGBM_DatasetGetFeatureNames,
        ds.handle => DatasetHandle,
        new_len => Cint,
        num_feature_names => Ref{Cint},
        new_buffer_len => Csize_t,
        out_buffer_len => Ref{Csize_t},
        feature_names => Ref{Ptr{UInt8}},
    )
```

src/wrapper.jl
```diff
@@ -482,24 +550,26 @@ end
 
 function LGBM_BoosterPredictForMat(bst::Booster, data::Matrix{T},
                                                               predict_type::Integer,
+                                                              start_iteration::Integer,
                                                               num_iteration::Integer,
                                                               is_row_major::Bool = false) where T<:Union{Float32,Float64}
     num_class = LGBM_BoosterGetNumClasses(bst)
-    lgbm_data_type = jltype_to_lgbmid(T)
+    data_type = jltype_to_lgbmid(T)
```
yalwan-iqvia, line 557: There wasn't a need to change this one, right? You don't need to change it back, it just looks to me like it wasn't necessary

test/basic_tests.jl
```diff
@@ -87,7 +87,7 @@ LGBM_PATH = if isabspath(LGBM_PATH) LGBM_PATH else abspath(joinpath(pwd(), "..",
     # Not the case while the random functions are enabled: bagging fraction and feature fraction
     # but they should generally be otherwise
     @test_broken isapprox(p_binary, p_multi[:, 2], rtol=1e-16)
```
yalwan-iqvia, line 89: I still can't see how the below works but these probabilities aren't similar. Is the order in p_multi the right way, i.e. is p_multi corresponding to probabilitiesfor 0, and 1 at indices 1,2 respectively? Is it reversed? Is it possibly random on the basis of the order of the occurrences in the training data?

test/ffi/base_utils.jl
```diff
@@ -10,7 +10,7 @@ using Test
 
     @test b != bb
     @test b !== bb
-    @test LightGBM.LGBM_BoosterSaveModelToString(b, 0, 0) == LightGBM.LGBM_BoosterSaveModelToString(bb, 0, 0)
+    @test LightGBM.LGBM_BoosterSaveModelToString(b, 0, 0, 0) == LightGBM.LGBM_BoosterSaveModelToString(bb, 0, 0, 0)
```
yalwan-iqvia, line 13: It might have been worth multiple dispatching these methods so you don't have to update every occurrence in the codebase

test/ffi/booster.jl
```diff
@@ -286,27 +290,50 @@ end
 
 @testset "LGBM_BoosterSaveModel" begin
```
yalwan-iqvia, line 291: nice, I would recommend cleaning up prior to running the test as well.

test/ffi/booster.jl
```diff
@@ -286,27 +290,50 @@ end
...
+    booster = LightGBM.LGBM_BoosterCreateFromModelfile(joinpath(@__DIR__, "data", "gain_test_booster"))
+    split_importance = LightGBM.LGBM_BoosterFeatureImportance(booster, 0, 0)
+    gain_importance = LightGBM.LGBM_BoosterFeatureImportance(booster, 0, 1)
+    split_save = LightGBM.LGBM_BoosterSaveModelToString(booster, 0, 0, 0)
+    gain_save = LightGBM.LGBM_BoosterSaveModelToString(booster, 0, 0, 1)
+    split_load = LightGBM.LGBM_BoosterLoadModelFromString(split_save)
+    gain_load = LightGBM.LGBM_BoosterLoadModelFromString(gain_save)
+    split_importance_loaded = LightGBM.LGBM_BoosterFeatureImportance(split_load, 0, 0)
+    gain_importance_loaded = LightGBM.LGBM_BoosterFeatureImportance(gain_load, 0, 1)
+    expected_gain = [89.73980999, 65.49232054, 112.80447054, 107.81817985, 124.81229973]
```
yalwan-iqvia, line 331: Why have you coded these values in here? The way I would imagine this test to work is as follows:

* load that model above (or any other)
* compute both its gain and split importance
* save the model twice, once with each of those gain and split parameters
* load each of the saved models to a new model
* for EACH model, compute gain and split importances, and show that both values are the same as the originals for both models

This then shows that we don't have to concern ourselves with which value we pass by default when using save model to string internally for serialisation/deserialisation purposes.

test/mlj/regression.jl
```diff
@@ -35,6 +35,9 @@ rmse_weights             = calc_rmse(y[test], yhat_with_weights)
 @test rmse < 0.05
 @test !isapprox(rmse, rmse_weights, rtol=0.1) # check that they differ by at least around 10% with/without weights
 
+# checking that fitting with and without weight outputs are non-identical
+@test yhat_with_weights != yhat
```
yalwan-iqvia, line 39: The test has been done differently by showing that the RMSE varies substantially, so we dont need this additional test

test/mlj/binary_classifier.jl
yalwan-iqvia: It would be interesting to check the raw prediction values. maybe the amount they vary is sufficiently small to not move actual predcitions (although this itself seems remarkably unlikely)

test/mlj/regression.jl
yalwan-iqvia: ( see two lines above)

kd-iqvia: Yes, I saw the test was done differently on this one. I added this part of troubleshooting the issue with the binary test. To make sure it wasn't a problem with multiclass, or regression. Makes sense to remove it

src/wrapper.jl
kd-iqvia: it was... C-API changed [see_changes](https://lightgbm.readthedocs.io/en/latest/C-API.html#c.LGBM_DatasetCreateFromMats) and this parameter got renamed which eventually caused segfaulting

test/mlj/regression.jl
kd-iqvia: 16ddb07 resolves

src/wrapper.jl
yalwan-iqvia: Are you sure it wasn't in connection with other changes? C-API doesn't understand (or see) parameter names. When I change it back manually there are no segfaults (and the dataset equivalent is still using the old variable name)

kd-iqvia: I can have another looksee but the segfault happened after enabling this test: https://github.com/IQVIA-ML/LightGBM.jl/blob/master/test/basic_tests.jl#L89 where to get the `p_binary`  and `p_multi` we're using the `predict` which then uses `LGBM_BoosterPredictForMat` . This function had API changes as they also added `start_iteration` but this still didn't help to resolve the segfault. After renaming the parameter to what they have in their API, I didn't get the segfault any more and could actually see the julia stacktrace for the test itself. But yes  While the order of the parameters has to be respected I was surprised that the name change would make a difference (which you're right it shouldn't). In fact I didn't change the param name elsewhere...

kd-iqvia: so yes it's not the `segfault` I was getting for missed API changes but I got the below error after I reversed this to be called `lgbm_data_type` and enabled the same test:
image

but after switching it back to `data_type` as in this PR, I actually got the error for the test itself before the full stacktrace and I didn't do any other changes:
image

test/ffi/booster.jl
kd-iqvia: 9c03935 resolves

src/wrapper.jl
kd-iqvia: OK... something dodgy must have been happening locally on my machine. I cleaned up Manifest.toml and then run both cases and yes the name of the parameter doesn't make any difference. I brought back the `lgbm_data_type`

src/utils.jl
```diff
@@ -42,10 +42,19 @@ Save the fitted model in `estimator` as `filename`.
 * `filename::String`: the name of the file to save the model in.
 * `num_iteration::Integer`: keyword argument that sets the number of iterations of the model that
     should be saved. `< 0` for all iterations.
+* `start_iteration` : : Start index of the iteration that should be saved.
+* `feature_importance_type` : Type of feature importance,
+    can be C_API_FEATURE_IMPORTANCE_SPLIT or C_API_FEATURE_IMPORTANCE_GAIN
 """
```
yalwan-iqvia, line 48: You probably want to add "average_precision" to the `MAXIMIZE_METRICS` constant above (and probably review it for any other maximisable metrics)

test/ffi/booster.jl
kd-iqvia: hopefully now it's better

test/basic_tests.jl
kd-iqvia: ok for this one, what would be a reasonably acceptable tolerance? I looked into the outputs and we're talking about the following: 
image

I was very generous with the tolerance...

yalwan-iqvia: I think you should look at statistics of this ... like is the difference nearly constant? does the direction vary (sometimes negative, sometime positive) etc. Because given the predictions are the same but probabilities are changed by 1%ish it looks like it could be a constant bias or something similar ...

test/mlj/binary_classifier.jl
kd-iqvia: somehow changing the number of samples works... but that's nasty

test/mlj/binary_classifier.jl
kd-iqvia: I looked at the raw prediction value for 1000k samples and yes they vary but somehow small. Changing the number of samples to other values somehow works but this looks suspicious.

kd-iqvia: I additionally ran several tests with v2.3.1 just to see if I can observe a similar behaviour with the various samples. Some conclusions:

-  on v2.3.1. this test fails but with relatively small samples (5-20) but is not reproducible with other numbers of samples
- on v3.0.0 this tests fails only with samples in anywhere <800 - 1001> (inclusive 800 and 1001) and works perfectly fine with anything below that (except for 5-20 just as it is with v2.3.1) and anything above that

test/ffi/base_utils.jl
kd-iqvia: isn't it better with defaults in this case as we showed in tests it doesn't matter which one is used so either the user specifies a value, or we go with a default

src/utils.jl
kd-iqvia: f045a87

test/basic_tests.jl
```diff
@@ -86,8 +86,9 @@ LGBM_PATH = if isabspath(LGBM_PATH) LGBM_PATH else abspath(joinpath(pwd(), "..",
     # check binary and multi are equivalent
     # Not the case while the random functions are enabled: bagging fraction and feature fraction
     # but they should generally be otherwise
-    @test_broken isapprox(p_binary, p_multi[:, 2], rtol=1e-16)
-    @test_broken binary_classes == multi_classes
+
+    @test isapprox(p_binary, p_multi[:, 2], rtol=1e1)
```
yalwan-iqvia, line 90: This is a huge tolerance, what are the statistics on `p_binary - p_multi[:, 2]` ?

test/ffi/booster.jl
```diff
@@ -286,27 +290,60 @@ end
...
+    split_load_model = LightGBM.LGBM_BoosterLoadModelFromString(split_model_save)
+    gain_load_model = LightGBM.LGBM_BoosterLoadModelFromString(gain_model_save)
+    # for EACH model, compute gain and split importances
+    split_importance_from_split_loaded_model = LightGBM.LGBM_BoosterFeatureImportance(split_load_model, 0, 0)
+    gain_importance_from_split_loaded_model= LightGBM.LGBM_BoosterFeatureImportance(split_load_model, 0, 1)
+    gain_importance_from_gain_loaded_model = LightGBM.LGBM_BoosterFeatureImportance(gain_load_model, 0, 1)
+    split_importance_from_gain_loaded_model = LightGBM.LGBM_BoosterFeatureImportance(gain_load_model, 0, 0)
+    # and show that both values are the same as the originals for both models
+    @test split_importance == split_importance_from_split_loaded_model
+    @test split_importance == split_importance_from_gain_loaded_model
```
yalwan-iqvia, line 344: hooray it worked :D

src/MLJInterface.jl
```diff
@@ -6,14 +6,13 @@ import LightGBM
...
-    "tweedie", "ndcg", "lambdarank", "map", "mean_average_precision", "auc", "binary_logloss", "binary",
-    "binary_error", "auc_mu", "multi_logloss", "multi_error", "cross_entropy", "xentropy", "multi_logloss",
-    "multiclass", "softmax", "multiclassova", "multiclass_ova", "ova", "ovr", "cross_entropy_lambda",
+    "tweedie", "ndcg", "lambdarank", "map", "mean_average_precision", "auc", "average_precision", "binary_logloss",
+    "binary_error", "auc_mu", "multi_logloss", "multi_error", "cross_entropy", "xentropy", "multi_logloss","cross_entropy_lambda",
     "xentlambda", "kullback_leibler", "kldiv",
 )
 
 const CLASSIFICATION_OBJECTIVES = (
-    "binary", "multiclass", "softmax", "multiclassova", "multiclass_ova", "ova", "ovr",
```
kd-iqvia, line 16: `multiclassova` and its aliases not needed for MLJInterface

src/MLJInterface.jl
```diff
@@ -6,14 +6,13 @@ import LightGBM
 
 const LGBM_METRICS = (
     "None", "l1", "l2", "rmse", "quantile", "mape", "huber", "fair", "poisson", "gamma", "gamma_deviance",
-    "tweedie", "ndcg", "lambdarank", "map", "mean_average_precision", "auc", "binary_logloss", "binary",
-    "binary_error", "auc_mu", "multi_logloss", "multi_error", "cross_entropy", "xentropy", "multi_logloss",
-    "multiclass", "softmax", "multiclassova", "multiclass_ova", "ova", "ovr", "cross_entropy_lambda",
```
kd-iqvia, line 11: minor cleanup as we had objectives listed in LGBM_METRICS

test/ffi/booster.jl
kd-iqvia: 🎉

yalwan-iqvia (approved):

PR #80, opened 2020-11-27
fit with pre-constructed dataset

chilledgeek-iqvia:
added fit method that takes in pre-constructed datasets

also removed num_iterations in the existing fit method as this was not really used anywhere (which makes it identical to the "old signature", so that can be removed as well)

### DIFF
--- a/src/fit.jl
+++ b/src/fit.jl
@@ -50,13 +50,10 @@ function fit!(
 
     test_dss = []
 
-    if length(test) > 0
-        log_debug(verbosity, "Started creating LGBM test datasets\n")
-        for test_entry in test
-            test_ds = LGBM_DatasetCreateFromMat(test_entry[1], ds_parameters, train_ds, is_row_major)
-            LGBM_DatasetSetField(test_ds, "label", test_entry[2])
-            push!(test_dss, test_ds)
-        end
+    for test_entry in test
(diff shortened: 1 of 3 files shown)


test/basic/fit.jl
```diff
@@ -0,0 +1,55 @@
...
+    binary_classes = LightGBM.predict_classes(estimator, train_matrix, verbosity = -1)
+
+    # Assert
+    # check also that binary outputs and classes are invariant in prediction output shapes, i.e. they're matrices
+    @test length(size(p_binary)) == 2
+    @test length(size(binary_classes)) == 2
+    
+end
+
+@testset "test fit with dataset as input type runs fine without test set -- binary" begin
```
yalwan-iqvia, line 34: Probably make the test names a little less verbose 😁

src/fit.jl
```diff
@@ -44,38 +48,44 @@ function fit!(
...
+        test_ds = LGBM_DatasetCreateFromMat(test_entry[1], ds_parameters, train_ds, is_row_major)
+        LGBM_DatasetSetField(test_ds, "label", test_entry[2])
+        push!(test_dss, test_ds)
+    end
+
+    return fit!(estimator, train_ds, test_dss..., verbosity=verbosity)
+end
+
+
+# Pass Dataset class directly to speed up data loading
```
yalwan-iqvia, line 63: specifically if the user has already constructed the dataset themselves and its part of an iterative process othwerwise it makes no difference

PR #86, opened 2020-12-04
fix early stopping and upgrade lightgbm version

chilledgeek:
This PR addresses:
- Bug with early stopping logic (+ tidying up)
- Upgrade lightgbm to v3.1.0 (which fixes some of the use of metrics)


src/fit.jl
```diff
@@ -89,41 +89,47 @@ end
 
 
 function train!(
-    estimator::LGBMEstimator, num_iterations::Int, tests_names::Vector{String}, verbosity::Integer, start_time::DateTime
+    estimator::LGBMEstimator, 
+    num_iterations::Int, 
+    tests_names::Vector{String}, 
+    verbosity::Integer, 
+    start_time::DateTime
```
yalwan-iqvia, line 96: trailing comma please

src/fit.jl
```diff
@@ -89,41 +89,47 @@ end
...
+            best_score[metric][tests_name] = -Inf
+            best_iter[metric][tests_name] = 1
+        end
+    end
 
     start_iter = get_iter_number(estimator) + 1
     end_iter = start_iter + num_iterations - 1
 
     metrics_idx_sequence = (((start_iter:end_iter) .- 1) .% estimator.metric_freq) .== 0
     total_metrics_evals = sum(metrics_idx_sequence)
```
yalwan-iqvia, line 118: Looks like this one can be deleted?

src/fit.jl
```diff
@@ -147,76 +153,72 @@ function eval_metrics!(
...
-                    best_iter[metric_idx, test_idx] = iter
-                elseif iter - best_iter[metric_idx, test_idx] >= estimator.early_stopping_round
-                    # This will shrink it up to the current stored metric
-                    shrinkresults!(results, metrics_store_idx)
-                    log_info(verbosity, "Early stopping at iteration ", iter,
-                             ", the best iteration round is ", best_iter[metric_idx, test_idx], "\n")
-                    return 1
-                end
+    # check early stopping condition
+    if estimator.early_stopping_round > 0
```
yalwan-iqvia, line 174: Is it right that metrics are grabbed at every iteration if early stopping is on at all?

src/fit.jl
```diff
@@ -147,76 +153,72 @@ function eval_metrics!(
...
+        for tests_name in tests_names
+            for (metric_idx, metric) in enumerate(metrics)
+                maximize_score = bigger_is_better[metric] * now_scores[tests_name][metric_idx]
+                # All good if maximize_score is better than previous best
+                if maximize_score > best_score[metric][tests_name]
+                    best_score[metric][tests_name] = maximize_score
+                    best_iter[metric][tests_name] = iter
+                    continue
+                # All good if difference between current and best iter is within early_stopping_round
+                elseif (iter - best_iter[metric][tests_name]) < estimator.early_stopping_round
```
yalwan-iqvia, line 184: wouldn't it be just the case that if iter < early_stopping_round we don't bail?

src/fit.jl
```diff
@@ -147,76 +153,72 @@ function eval_metrics!(
...
-        if !haskey(results, evalname)
-            results[evalname] = Dict{String,Vector{Float64}}()
-            results[evalname][metric_name] = Array{Float64}(undef,num_evals)
-        elseif !haskey(results[evalname], metric_name)
-            results[evalname][metric_name] = Array{Float64}(undef,num_evals)
-        end
-        results[evalname][metric_name][store_idx] = scores[metric_idx]
+    if !haskey(results, dataset_key)
+        results[dataset_key] = Dict{String,Vector{Float64}}()
+        results[dataset_key][metric_name] = Vector{Float64}()
```
yalwan-iqvia, line 217: A short hand for initialisaing an entry Vector of type Bla is to do `Bla[]`

src/fit.jl
```diff
@@ -147,76 +153,72 @@ function eval_metrics!(
...
     results::Dict{String,Dict{String,Vector{Float64}}},
-    estimator::LGBMEstimator,
-    store_idx::Integer,
-    num_evals::Integer,
-    evalname::String,
-    scores::Vector{Cdouble},
-    metrics::Vector{String},
+    dataset_key::String,
+    metric_name::String,
+    value_to_add::Real,
```
yalwan-iqvia, line 213: This type is a bit odd given that everywhere else it clearly says its F64

src/fit.jl
chilledgeek: I thought so at first, but I think it does not cover all of the scenarios. Using this sample:

Sample run results:
```
iterations, metric (e.g. auc)
1, 0.4
2, 0.5
3, 0.5
4, 0.6
5, 0.7
6, 0.6
7, 0.7
8, 0.4
9, 0.9
10, 0.8
```
- For these test cases:
  - if early_stopping_round <= 0, run should not stop till the end (so get 0.8)
    - Covered (as we first check `early_stopping_round > 0`)
- However for these cases:
  - if early_stopping_round == 2, run should stop at round 5 (rounds 6 and 7 don't show improvement)
    - If we use `iter` only, when `iter == 6`, early_stopping_round will kick in immediately as `iter < early_stopping_round ` (`6 < 2`) and it won't even consider round 7? So in this case it is a coincidence that the results are as expected...
  - if early_stopping_round == 4, run should not stop till the end (continue to see improvement every 4 iterations till the end)
    - Again if we use just `iter`, then this run will stop at round 4, which is what we don't want. Instead, if we update and compare the difference between `iter` and the `best_iter` round to `early_stopping_round`, then that seems correct to me.

Let me know if i'm missing anything obvious! :D

src/fit.jl
chilledgeek: Yes! Thanks! (I was aiming to refactor the code so that I could get rid of it...but forgot to do so after all the refactoring!) :D

src/fit.jl
chilledgeek: that would depend on the criteria `(iter - 1) % estimator.metric_freq == 0` in line 161, right?  If that condition is not met, now_scores will be an empty dict, and nothing will really be stored/printed

src/fit.jl
yalwan-iqvia: Ok yes I can see it follows the logic of the LightGBM callback here: https://github.com/microsoft/LightGBM/blob/37da4ed755897ba16e065eecf18e31d03927e91b/python-package/lightgbm/callback.py#L236

So the way it works is that it needs to see an improvement for at least early_stopping_round iterations to continue training

src/MLJInterface.jl
```diff
@@ -216,7 +216,7 @@ function fit(mlj_model::MODELS, verbosity::Int, X, y, w=AbstractFloat[])
     X = MLJModelInterface.matrix(X)
     # The FFI wrapper wants Float32 for these
     w = Float32.(w)
-    train_metrics = LightGBM.fit!(model, X, y_lgbm; verbosity=verbosity, weights=w)
+    train_metrics = LightGBM.fit!(model, X, y_lgbm[:]; verbosity=verbosity, weights=w)
```
yalwan-iqvia, line 219: I know why this fixes it but it isn't necessarily obvious to the casual reader: perhaps put a comment explaining that slicing notation creates copies and LGBM needs contiguous in memory data?

yalwan-iqvia (approved):
LGTM when address final comment

src/MLJInterface.jl
chilledgeek: done :)

PR #94, opened 2021-01-20
correct predict output shape

chilledgeek:
- Changed `loadmodel` to `loadmodel!` as we are actually modifying the booster with this method
- Currently if we use predict (in /src/predict.jl) on a binary classifier, the output is split into two columns while we only expected it to give one. The reason is that num_class is not loaded when we do `loadmodel!`. So loading in the num_class property when we `loadmodel!` fixes the predict functionality. Have also added tests on this

### DIFF
--- a/src/LightGBM.jl
+++ b/src/LightGBM.jl
@@ -18,7 +18,7 @@ include("cv.jl")
 include("search_cv.jl")
 include(joinpath(@__DIR__, "MLJInterface.jl"))
 
-export fit!, predict, predict_classes, cv, search_cv, savemodel, loadmodel
+export fit!, predict, predict_classes, cv, search_cv, savemodel, loadmodel!
 export LGBMEstimator, LGBMRegression, LGBMClassification
 
 end # module LightGBM
--- a/test/basic/test_utils.jl
+++ b/test/basic/test_utils.jl
@@ -1,4 +1,9 @@
+using Test
 using LightGBM
+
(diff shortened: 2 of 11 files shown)


src/LightGBM.jl
```diff
@@ -18,7 +18,7 @@ include("cv.jl")
 include("search_cv.jl")
 include(joinpath(@__DIR__, "MLJInterface.jl"))
 
-export fit!, predict, predict_classes, cv, search_cv, savemodel, loadmodel
+export fit!, predict, predict_classes, cv, search_cv, savemodel, loadmodel!
```
yalwan-iqvia, line 21: Nicely done, I've messed this up before.

test/basic/test_utils.jl
```diff
@@ -8,3 +13,49 @@ using LightGBM
     expected = "categorical_feature=0,2,4,6,8"
     @test occursin(expected, ds_parameters)
 end
+
+
+
+@testset "loadmodel predicts same as original model -- regression" begin
```
yalwan-iqvia, line 19: Nice tests. To make this more rigorous, I would  load the model onto a newly constructed default estimator and show predictions come out the same. Maybe use more non-default params for the original estimator too (so we can in fact spot/verify that predictions are invariant to the other params)

chilledgeek:
incremented version as well for new patch release

yalwan-iqvia (approved):

PR #96, opened 2021-01-29
Extend compat MLJModelInterface = "^0.3.6,^0.4"

ablaom:


yalwan-iqvia:
Any effects of this or is it bringing in MLJ2?

ablaom:
This should have no effect at your end. It is needed because of some code re-organization between repos (including new StatisticalTraits.jll package, now a tiny dep of MLJModelInterface). 

"mlj2" is already provided for in MLJModelInterface 0.3.7. However, to benefit from any "mlj2" buy-ins,  MLJ users will need MLJBase 0.17.1 (which will require MLJModelInterface 0.4, for unrelated reasons) but MLJ itself is not yet synched. This will happen after MLJModels catches up, next  week I hope.

Rolling out breaking releases from the bottom up is painful but I'm getting there. Also working on some changes to the model registry to make it more robust to changes upstream. So, fingers crossed, this won't happen again until MLJModelInterface 1.0 😉  

Did I answer your question?

yalwan-iqvia:
Thank you!

PR #97, opened 2021-01-29
Access the best iteration from early stopping

chilledgeek:
This PR addresses some issues related early stopping:

- The `best_iter` is a new key in the output of `fit!`/`train!` that captures the best iteration if early stopping is enabled (also added to the MLJ report)
- The original output is nested inside a dict key "metrics"
- We have added a `truncate_booster` Boolean flag (defaults to true) in `fit!`/`train!` that determines whether or not a model should be truncated to the best_iteration
- For running in MLJ, the `truncate_booster` flag can be declared when constructing the `LightGBM.MLJInterface.LGBMClassifier`, which will be passed to the `fit!`/`train!` methods when called

### DIFF
--- a/src/MLJInterface.jl
+++ b/src/MLJInterface.jl
@@ -21,6 +21,9 @@ const REGRESSION_OBJECTIVES = (
     "mean_absolute_percentage_error", "gamma", "tweedie",
 )
 
+const NON_LIGHTGBM_PARAMETERS = (
+    :truncate_booster,
+)
 
 MLJModelInterface.@mlj_model mutable struct LGBMRegressor <: MLJModelInterface.Deterministic
 
@@ -181,7 +184,7 @@ function mlj_to_kwargs(model::MODELS)
     return Dict{Symbol, Any}(
         name => getfield(model, name)
         for name in fieldnames(typeof(model))
-        if name != :truncate_booster
+        if !(name in NON_LIGHTGBM_PARAMETERS)
     )
 
 end
@@ -226,7 +229,6 @@ function fit(mlj_model::MODELS, verbosity::Int, X, y, w=AbstractFloat[])
     cache = (
         num_boostings_done=[LightGBM.get_iter_number(model)], 
         training_metrics=train_results["metrics"], 
-        best_iter=train_results["best_iter"],
     )
     report = user_fitreport(model, train_results)
 
@@ -267,7 +269,14 @@ function update(mlj_model::MLJInterface.MODELS, verbosity::Int, fitresult, cache
     old_lgbm_model.num_iterations = num_iterations + additional_iterations
 
     # eh this is ugly, possibly prompts a need for some refactoring
-    train_results = LightGBM.train!(old_lgbm_model, additional_iterations, String[], verbosity, LightGBM.Dates.now(), truncate_booster=old_mlj_model.truncate_booster)
+    train_results = LightGBM.train!(
+        old_lgbm_model, 
+        additional_iterations, 
+        String[], 
+        verbosity, 
+        LightGBM.Dates.now(); 
+        truncate_booster=old_mlj_model.truncate_booster
+    )
     fitresult = (old_lgbm_model, old_classes, deepcopy(mlj_model))
 
     final_num_iter = LightGBM.get_iter_number(old_lgbm_model)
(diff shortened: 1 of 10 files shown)


src/MLJInterface.jl
```diff
@@ -167,18 +168,20 @@ MLJModelInterface.@mlj_model mutable struct LGBMClassifier <: MLJModelInterface.
...
 MODELS = Union{LGBMClassifier, LGBMRegressor}
 
 
-function mlj_to_kwargs(model::LGBMRegressor)
+function mlj_to_kwargs(model::MODELS)
 
     return Dict{Symbol, Any}(
         name => getfield(model, name)
         for name in fieldnames(typeof(model))
+        if name != :truncate_booster
```
yalwan-iqvia, line 184: We might end up adding a few more things here. Maybe get a constant somewhere to keep track of these things?

src/MLJInterface.jl
```diff
@@ -195,11 +198,7 @@ function mlj_to_kwargs(model::LGBMClassifier, classes)
         num_class = 1
     end
 
-    retval = Dict{Symbol, Any}(
```
yalwan-iqvia, line 198: Thanks for killing this dupe

src/MLJInterface.jl
```diff
@@ -264,14 +267,19 @@ function update(mlj_model::MLJInterface.MODELS, verbosity::Int, fitresult, cache
     old_lgbm_model.num_iterations = num_iterations + additional_iterations
 
     # eh this is ugly, possibly prompts a need for some refactoring
-    train_metrics = LightGBM.train!(old_lgbm_model, additional_iterations, String[], verbosity, LightGBM.Dates.now())
```
yalwan-iqvia, line 267: Its clearly not necessary, but perhaps use a `;` to separate kwargs

src/MLJInterface.jl
```diff
@@ -264,14 +267,19 @@ function update(mlj_model::MLJInterface.MODELS, verbosity::Int, fitresult, cache
...
     iteration_history = deepcopy(cache.num_boostings_done)
     push!(iteration_history, final_num_iter - num_iterations)
-    report = user_fitreport(old_lgbm_model, cache.training_metrics, train_metrics)
-    newcache = (num_boostings_done=iteration_history, training_metrics=report.training_metrics)
+
+    report = user_fitreport(old_lgbm_model, cache.training_metrics, train_results)
+    newcache = (
+        num_boostings_done=iteration_history, 
+        training_metrics=report.training_metrics,
+        best_iter = report.best_iter,
```
yalwan-iqvia, line 281: Why does this need to be in the cache?

src/fit.jl
```diff
@@ -132,28 +139,44 @@ function train!(
...
 )
-    return train!(estimator, estimator.num_iterations, tests_names, verbosity, start_time)
+    return train!(estimator, estimator.num_iterations, tests_names, verbosity, start_time, truncate_booster=truncate_booster)
+end
+
+
+function truncate_model!(estimator::LGBMEstimator, best_iteration::Integer)
+    current_iteration = LGBM_BoosterGetCurrentIteration(estimator.booster)
+    times_to_rollback = current_iteration - best_iteration # current_iteration must be >= best_iteration
+    for _ in [1:times_to_rollback;]
```
yalwan-iqvia, line 164: I think you can just iterate over the 1:n range syntax without wrapping it up into an array

test/ffi/booster.jl
```diff
@@ -163,8 +163,23 @@ end
...
+    booster = LightGBM.LGBM_BoosterCreate(dataset, verbosity)
+
+    # update learning 20 times
+    for _ in [1:20;]
+        finished = LightGBM.LGBM_BoosterUpdateOneIter(booster)
+    end
+
+    # Act and Assert
+    for n in reverse([1:20;])
+        @test LightGBM.LGBM_BoosterGetCurrentIteration(booster) == n
```
yalwan-iqvia, line 180: Here I'd like to propose extension of this test:

use savemodeltostring to save the serialised form and also check that the num_trees property matches expected value

src/MLJInterface.jl
chilledgeek: you're right it doesn't. Will remove this. thanks!

test/ffi/booster.jl
```diff
@@ -163,8 +163,32 @@ end
...
+
+        model_string = LightGBM.LGBM_BoosterSaveModelToString(booster)
+        tree_sizes_in_string = match(r"\ntree_sizes=(.*)\n", model_string)[1]
```
yalwan-iqvia, line 183: Thats good just put a comment to explain it

chilledgeek: Done :)

yalwan-iqvia (approved):
LGTM

PR #101, opened 2021-02-09
updated defaults on min_sum_hessian_in_leaf

kd-iqvia:
updated default to match docs: https://lightgbm.readthedocs.io/en/latest/Parameters.html#min_sum_hessian_in_leaf


yalwan-iqvia (approved):
LGTM

PR #103, opened 2021-02-11
Fix parameter passing

yalwan-iqvia:
It seems like some parameters were not being passed to dataset or booster.

Rather than trying to fix this by hand, I took advantage of the fact that LightGBM doesn't seem to warn if you pass all the parameters to both constructs, so I updated `stringifyparams` to just ignore the `booster` and `model` fields

I also took the opportunity to give it a kwarg to add verbosity and fix verbosity passing everywhere -- finally tests don't vomit output everywhere.

### DIFF
--- a/src/cv.jl
+++ b/src/cv.jl
@@ -22,9 +22,12 @@ last valid iteration.
 * `verbosity::Integer`: keyword argument that controls LightGBM's verbosity. `< 0` for fatal logs
     only, `0` includes warning logs, `1` includes info logs, and `> 1` includes debug logs.
 """
-function cv(estimator::LGBMEstimator, X::Matrix{TX}, y::Vector{Ty}, splits;
-                               verbosity::Integer = 1, truncate_booster::Bool = true) where {TX<:Real,Ty<:Real}
-    ds_parameters = stringifyparams(estimator, DATASETPARAMS)
+function cv(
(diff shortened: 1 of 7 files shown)


yalwan-iqvia:
@chilledgeek @kd-iqvia PTAL

src/cv.jl
```diff
@@ -22,9 +22,12 @@ last valid iteration.
...
 """
-function cv(estimator::LGBMEstimator, X::Matrix{TX}, y::Vector{Ty}, splits;
-                               verbosity::Integer = 1, truncate_booster::Bool = true) where {TX<:Real,Ty<:Real}
-    ds_parameters = stringifyparams(estimator, DATASETPARAMS)
+function cv(
+    estimator::LGBMEstimator, X::Matrix{TX}, y::Vector{Ty}, splits;
+    verbosity::Integer = 1, truncate_booster::Bool = true,
+) where {TX<:Real,Ty<:Real}
+
+    ds_parameters = stringifyparams(estimator; verbosity=verbosity)
```
kd-iqvia, line 30: I think this is a wiser approach as it reduces the chances of missing or wrongly allocating params and hopefully is futureproof

kd-iqvia (approved):
Thanks for doing that and sorting out verbosity too. I think this implementation choice reduces the chances of missing or misplacing the parameters allocations in to the right "bucket" and lets LightGBM to do the right thing.

PR #104, opened 2021-02-18
Add "Custom LightGBM binaries" item to document

wakakusa:
Added "Custom LightGBM binaries" from Readme.md that was not on Document


yalwan-iqvia:
Thank you very much @wakakusa 

Can you comment on if there is a nice way to synchronise README and this so that it doesn't occur in future?

I won't block this PR on the change,

PR #106, opened 2021-04-06
added more GPU params and upgraded LightGBM

kd-iqvia:
- added more GPU parameters to enable GPU tuning 
- upgraded LightGBM to v3.2.0

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@
 # This is also pinned to some internal hackery for travis windows setup
 language: julia
 before_install:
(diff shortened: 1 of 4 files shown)


chilledgeek:
would be good to have tests to check the params work, but i guess we don't have the hardware to actually check!

so, LGTM :)

chilledgeek (approved):

kd-iqvia:
> would be good to have tests to check the params work, but i guess we don't have the hardware to actually check!
> 
> so, LGTM :)

yeah... for the GPU params it's hard to add unit tests but no worries they will be used to tune some real GPU runs 😄

Issue #105, opened 2021-03-05
LD_LIBRARY_PATH gets overrode by pre-compiled lightGBM.jl binary


kd-iqvia:
I've encountered a possible bug when running lightGBM.jl with my custom binary. Despite of pointing my LD_LIBRARY_PATH to the custom binary, whatever is existing in here: `~/.julia/compiled/$JULIA_VERSION/LightGBM` will be used.
The current workaround is to: `rm -rf ~/.julia/compiled/v1.4/LightGBM` and then the exported LD_LIBRARY_PATH will work correctly.
Ideally, we'd like to avoid doing it each time.

PR #108, opened 2021-04-16
fixing precompile caching of the library path

yalwan-sage:
Addresses #105 

Tested locally by re-invoking julia many times and seeing that the log message appears each time, and also by setting LD_LIBRARY_PATH and observing different log message appearing


kd-iqvia:
thanks @yalwan-sage for your contribution and sorting this out!

kd-iqvia:
sorry Friday brain (or the 😺 pressed the button)

PR #113, opened 2021-08-26
Adding support for constructing Dataset from SparseMatrix 

yaxxie:
As title:

- [x] Implement FFI wrapper for creating sparse dataset
- [x] Tests for FFI
- [x] Add support to `fit!`
- [x] Tests for higher level functions
    - [x] Specifically tests to show that the SparseMatrix support is fitting models identically with otherwise identical data
- [ ] [Optional?] Figure out what needs to be done for MLJ
- [ ] [Optional?] Add support for SparseMatrix predictions?

### DIFF
--- a/test/basic/test_fit.jl
+++ b/test/basic/test_fit.jl
@@ -3,9 +3,11 @@ module TestFit
 using Test
 using Dates
 using LightGBM
+using SparseArrays
 
 # test fixtures
 train_matrix = rand(5000,70) # create random dataset
+train_sparse = sparse(train_matrix)
 train_labels = rand([0, 1], 5000)
 train_dataset = LightGBM.LGBM_DatasetCreateFromMat(train_matrix, "")
 LightGBM.LGBM_DatasetSetField(train_dataset, "label", train_labels)
@@ -24,7 +26,7 @@ LightGBM.LGBM_DatasetSetField(test2_dataset, "label", test2_labels)
 @testset "test fit! with dataset -- binary" begin
     # Arrange
     estimator = LightGBM.LGBMClassification(objective = "binary", num_class = 1)
-    
+
     # Act
     LightGBM.fit!(estimator, train_dataset, test_dataset, verbosity = -1);
     p_binary = LightGBM.predict(estimator, train_matrix, verbosity = -1)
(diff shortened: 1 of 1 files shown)


yaxxie:
@kd-iqvia do you consider this sufficient for an initial take for sparse matrix support?

The tests aren't exhaustive (e.g. I don't demonstrate that you can in fact mix dense/sparse matrices for the train test, and I haven't covered a huge range of scenarios for checking the sparse inputs are correctly received), but it looks like it works. Is there anything you want to see or would need to see before merge?

src/wrapper.jl
```diff
@@ -110,9 +128,40 @@ macro lightgbm(f, params...)
...
+    idx_type, data_type = sparsetypes(data)
+
+    out = Ref{DatasetHandle}()
+    @lightgbm(
+        :LGBM_DatasetCreateFromCSC,
+        data.colptr .- 1 => Ptr{Nothing},
+        idx_type => Cint,
+        Int32.(data.rowval .- 1) => Ptr{Cint},
+        data.nzval => Ptr{Nothing},
+        data_type => Cint,
```
FatemehTahavori, line 154: data_type – Type of data pointer, can be C_API_DTYPE_FLOAT32 or C_API_DTYPE_FLOAT64 shouldn't be this type?

src/wrapper.jl
```diff
@@ -5,6 +5,8 @@ const C_API_DTYPE_FLOAT32 = 0
 const C_API_DTYPE_FLOAT64 = 1
 const C_API_DTYPE_INT32 = 2
 const C_API_DTYPE_INT64 = 3
+const C_API_MATRIX_TYPE_CSC = 1
+const C_API_MATRIX_TYPE_CSR = 0
```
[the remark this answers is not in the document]
yaxxie: Not for this PR but they would get used for the other sparse support API calls, in particular PredictSparse*

src/wrapper.jl
yaxxie: `data_type` is a variable, which is one of `C_API_DTYPE_FLOAT32` `C_API_DTYPE_FLOAT64` (determined by `sparsedatatype` call, see https://github.com/IQVIA-ML/LightGBM.jl/pull/113/files#diff-e08580f7ddadf1983cf458c25618ce250ee72b50d5924cd1375df3c4b023d963R102) 
`Cint` is the type that needs to be passed to the FFI inteface (under the hood, `ccall`) 

So what we're doing here is passing a C integer, which takes a value encoding F32 or F64, and telling `ccall` that this variable is a C integer

Issue #78, opened 2020-11-11
possible bug with MLJInterface


kainkad:
While upgrading LightGBM to v.3.0.0 an issue has been found with one of the previously passing tests now failing: https://github.com/IQVIA-ML/LightGBM.jl/blob/master/test/mlj/binary_classifier.jl#L41 where we compare predictions for models fitted with and without weights. The issue has been troubleshooted and indicated a possible bug with the MLJInterface.
The below code shows that the issue wasn't with the LightGBM upgrade itself as the same test has passed after removing the MLJInterface and fitting and predicting directly with LightGBM:

```julia
using MLJBase
using Test
using Random: seed!

import CategoricalArrays
import LightGBM

N = 2
seed!(0)

X       = rand(1000, 3)
ycat    = string.(mod.(round.(Int, X[:, 1] * 10), N)) |> MLJBase.categorical
weights = Float64.(MLJBase.int(ycat)) # just use the 1's/2's directly as multipliers

y = MLJBase.int(ycat) .-1 # make plain Vector with categ. elements (actually not sure what this is doing)


train, test = MLJBase.partition(MLJBase.eachindex(y), 0.6) # prepare train and test data
X_train = MLJBase.selectrows(X, train)
y_train = y[train]
X_test = MLJBase.selectrows(X, test)


model = LightGBM.LGBMClassification(objective="binary", num_class=1, num_iterations=100)
LightGBM.fit!(model, X_train, y_train, verbosity = -1) # fit once, without weights
yhat = LightGBM.predict(model, X_test, verbosity = -1)


weights_train = Float32.(weights[train])
model_weights = LightGBM.LGBMClassification(objective="binary", num_class=1, num_iterations=100)#the below works if we just used model again but wanted to make sure we're not overriding anything
LightGBM.fit!(model_weights, X_train, y_train; weights=weights_train, verbosity = -1) # fit again with weights
yhat_with_weights = LightGBM.predict(model_weights, X_test, verbosity = -1)


@test yhat_with_weights != yhat # All we can really say about fitting with/without weights for this example is that the solutions shouldn't be identical
```

yalwan-iqvia:
When we make a new release (soon-ish) with 3.1 pre-bundled, we should probably revisit this in conjunction with upgraded MLJ libs and see if the problem disappears or not.

kd-iqvia:
[...]

PR #119, opened 2022-05-16
Update to latest bundled binaries

yaxxie:
- [x] Fix the CI
- [x] Add 1.7 to CI matrix
- [x] Fix 1.7 tests
- [x] Upgrade bundled binary to latest LGBM release (3.3.2 as of writing)
- [x] Update README


yaxxie:
@IQVIA-ML/lightgbm-maintainers 
As you can see, the CI for 1.0, 1.1 and 1.2 is failing.
It is just dependency resolution for lower versions (specifically related to MLJ)
Either I can drop CI coverage for these lower versions, or break tests up so that MLJ tests are run for upper versions and regular tests for lower versions.

If we see [here](https://discourse.julialang.org/t/julia-v1-7-0-has-been-released-and-v1-6-is-now-lts/72336) we can see that 1.6.x is the new LTS (supplanting 1.0.x series).

So the question becomes if you're ok with dropping official support for pre-1.6 releases.

yaxxie:
P.S. I'm working on 1.7 CI fixes.
P.P.S I'm also going to upgrade the bundled binary

yaxxie:
@ablaom I'm doing a bit of investigation as to whats happening for our CI and I found that for v1.0, 1.1 and 1.2 julia versions 

If I have `MLJModelInterface@1.4.2` and ask for `MLJBase` the solver gives MLJBase@0.10.1, which is completely wild. In the meantime If I just ask for `MLJModelInterface` and `MLJBase` the solver gives much more reasonable versions of `1.3.3` and `0.18.19` respectively.

Because I'd like a) the CI to not fail when it can pass (and I can't actually find someone to turn off early 1.x as required checks) and b) I personally find it quite unsatisfying that we can't continue to support the early 1.x series, I'm going to pin this package to the `1.3.x` series of `MLJModelInterface` -- I'd be grateful if you could summarise what issues this might cause for us down the line.

**IFF** you also have the inclination, it'd be nice to see this working again. I appreciate that the early 1.x series will never any longer get cutting edge MLJ packages but it doesn't make sense for the solver to give versions which are prior to the initial release of this LGBM package. I totally appreciate you may not care for this given that 1.6.x is now LTS, but as I mentioned, I'd personally appreciate it.

yaxxie:
@ablaom as a follow up, I can see that pre-1.6 won't be supported. I still think the wonky package resolution is weird but since MLJ no longer cares about pre-1.6 users I've made it so our tests only operate on 1.6+

ablaom:
I agree it would be nice to maintain support for < 1.6. However, there is very little appetite for this among developers, probably because we are all spread so thin. I recall that the SciML organisation was actively discouraging devs to support < 1.6.  

MLJModelInterface 1.4 adds a couple of traits and some doc-string convenience functions. Adding traits should not be breaking - we took special precautions. So I don't see why MLJModelInterface 1.3 is a problem if you want to use an old Julia version. But, as you can see, we've given up and no longer support Julia 1.5 in MLJBase.

ablaom:
FYI: [We are rolling out](https://github.com/alan-turing-institute/MLJ.jl/issues/913https://github.com/alan-turing-institute/MLJ.jl/issues/9130) standardized docstrings for all MLJ models. I have a technical writer that may be make a PR, but if you're inclined to do this yourself, all the better.

.github/workflows/main.yml
```diff
@@ -39,18 +41,35 @@ jobs:
...
+      - name: v1.2 hackfix
+        if: matrix.version == '1.2'
+        run: rm test/Project.toml
       - run: julia --color=yes --project -e 'using Pkg; Pkg.build(); Pkg.test(coverage=true);'
         continue-on-error: false
         env:
           LIGHTGBM_EXAMPLES_PATH: LightGBM-2.3.1
+  MLJ:
+    name: MLJ Interface enabled -- Julia ${{ matrix.version }}
+    runs-on: ubuntu-latest
```
kd-iqvia, line 53: @yaxxie just wondering if we should add at least mac-OS as well given there are quite a few mac users of this package, ideally all 3. I think it should work with the `matrix:` and then `os:  [windows-latest, ubuntu-latest, macOS-latest]` unless you had a specific reason why you went for `runs-on`? It will add few more runs to the CI but at least we'll sort of know that the supported versions with MLJ work across platforms.

test/mlj/binary_classifier.jl
```diff
@@ -14,30 +13,33 @@ model = LightGBM.MLJInterface.LGBMClassifier(objective="binary", num_iterations=
 
 # test binary case:
 N = 2
-seed!(0)
+Nsamples = 3000
 
-X       = (x1=rand(1010), x2=rand(1010), x3=rand(1010))
```
kd-iqvia, line 19: does this resolve #78 ?

kd-iqvia:
Thanks @yaxxie for this PR and yes splitting the CI tests in this case to get all versions run for lightgbm and 1.6+ with MLJ makes sense given the comments from @ablaom 
I just added minor comments but LGTM

kd-iqvia:
@ablaom thanks for the heads up on the standardised docstrings. I couldn't open the URL but have looked at the docstrings for MLJ models. A PR would be helpful

test/mlj/binary_classifier.jl
yaxxie: Probably not 😅😅😅 I didn't investigate

.github/workflows/main.yml
yaxxie: So the short answer ia that there's no need.

The reason why is that the FFI interface is really the only thing that needs to be tested cross platform, for the rest of the code theres no reason to believe there would be platform specific behaviour. I think this is acutely true for the MLJ interface, which is a much higher level.

yaxxie:
I have a few things to fix which @FatemehTahavori pointed out to me so please hold off merging

FatemehTahavori (approved):

ablaom:
@yaxxie Sorry, here's the doc-roll out link: link : https://github.com/alan-turing-institute/MLJ.jl/issues/913
The new standard is detailed here: https://alan-turing-institute.github.io/MLJ.jl/dev/adding_models_for_general_use/#Document-strings

PR #124, opened 2022-07-28
Fix download the binaries

FatemehTahavori:


FatemehTahavori:
This change is not needed as the src file is already there.

PR #126, opened 2022-08-23
Fix failing tests

FatemehTahavori:
at the moment master is failing. This MR fixing those tests.

### DIFF
--- a/test/ffi/booster.jl
+++ b/test/ffi/booster.jl
@@ -382,7 +382,7 @@ end
     string_repr = LightGBM.LGBM_BoosterSaveModelToString(booster)
     # so it turns out that the string save and file save aren't necesarily the same so..
     # check a bunch of expected substrings, etc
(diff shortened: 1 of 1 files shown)


yaxxie:
I'm not confortable with adding compat just to make tests pass (which otherwise pass).
If theres a problem within a specific docker image or environment, and its just to do with tests, considering ignoring the LGBM tests or configuring the environment properly.

FatemehTahavori:
The reason why I added this MR was to fix these tests. For me locally these tests are failing. I don't understand why these tests are passing in CI tests not locally. I will close this MR to first reproduce this issue

yalwan-sage:
Which platform? If its mac m1 its likely to be to do with mixed architecture libs and your local julia installation being of a different architecture to the standard libs. If you're using intel julia, you could try getting intel python (libpython comes with python -- using the rosetta brew install will probably work)  and making sure it is in the system linker path.

PR #128, opened 2022-08-23
Don't try find_library tests if the system doesn't satisfy the assumptions

yaxxie:
Recent M1 (cross architecture libs) issues and a weird docker image highlight a few circumstances where the assumed-to-be-present libs for the `find_library` tests aren't always present (in the expected forms, i.e. `libname.system_extension`)

To stop this test from failing and causing issues with users, if no such lib is found, the system is declared "broken" and these tests will subsequently be skipped as "broken", which doesn't cause alarming test failure messages.

### DIFF
--- a/test/basic/test_lightgbm.jl
+++ b/test/basic/test_lightgbm.jl
@@ -17,6 +17,8 @@ function setup_env()
     loaded = Libdl.dllist()
     # get all the ones which just have the system extension and no funny business
     withext = loaded[endswith.(loaded, Ref(Libdl.dlext))]
+    # Some versions of LLVM are ... weird ... in that they don't like being double loaded. Use something else
+    withext = withext[.!occursin.(Ref("LLVM"), withext)]
     # get the libname without prefix and extension
     withoutext = libname.(withext)
     # check which are loadable (or findable by find_library, this is WILD if they're not all found)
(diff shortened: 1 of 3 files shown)


test/basic/test_lightgbm.jl
```diff
@@ -44,37 +61,49 @@ end
     @testset "find_library works with no system lib" begin
 
         # Arrange
-        settings = setup_env()
-        cp(settings["ref_lib_lightgbm_path"], settings["lib_not_on_sys_fixture_path"]) # fake file copied from lightgbm
+        settings, isborked = setup_env()
+        if !isborked
+            cp(settings["linkable_path"], settings["lib_not_on_sys_fixture_path"], force=true, follow_symlinks=true)
```
danielsoutar, line 66: This and its counterpart in the test below seems to be the offending lines for the macOS CI failures:
```
 IOError: open("/usr/lib/libSystem.B.dylib", 0, 0): no such file or directory (ENOENT)
```

Looking at the Apple developer forums, they mention [`dlopen` code potentially failing if file existence is checked](https://developer.apple.com/forums/thread/655588).

yaxxie:
@danielsoutar I've reverted these tests to the basic form. Now it will detect if its a broken system and just show these tests as "broken", which won't cause alarming failures. In addition, I refactored it to run setup only once, and removed a redundant test that wasn't checking anything new.

Issue #131, opened 2023-05-01
Bug: categorical_feature is mutated inside stringifyparams


characat0:
The field `LGBMEstimator.categorical_feature` is mutated inside stringifyparams.
```julia
julia> import LightGBM

julia> classifier = LightGBM.LGBMClassification(categorical_feature = [1, 3, 5, 7, 9]);

julia> LightGBM.stringifyparams(classifier; verbosity=-1);

julia> println(classifier.categorical_feature)
[0, 2, 4, 6, 8]

julia> LightGBM.stringifyparams(classifier; verbosity=-1);

julia> println(classifier.categorical_feature)
[-1, 1, 3, 5, 7]
```

PR #133, opened 2023-05-06
fix: prevent mutation of categorical_feature inside stringifyparams

characat0:
Fixes https://github.com/IQVIA-ML/LightGBM.jl/issues/131

Assign to a new variable instead of assigning inplace. This is useful during CV where the model is trained many times.

----
This is a copy of https://github.com/IQVIA-ML/LightGBM.jl/pull/132 but from a cleaner branch

### DIFF
--- a/test/basic/test_utils.jl
+++ b/test/basic/test_utils.jl
@@ -17,6 +17,16 @@ y_train_regression = rand(1000)
 end
 
 
+@testset "stringifyparams -- multiple calls won't mutate fields" begin
+    indices = [1, 3, 5, 7, 9]
+    classifier = LightGBM.LGBMClassification(categorical_feature = indices)
+    LightGBM.stringifyparams(classifier; verbosity=-1)
+    ds_parameters = LightGBM.stringifyparams(classifier; verbosity=-1)
+
+    expected = "categorical_feature=0,2,4,6,8"
+    @test occursin(expected, ds_parameters)
+end
+
 
 @testset "loadmodel predicts same as original model -- regression" begin
     # Arrange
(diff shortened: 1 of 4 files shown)


test/basic/test_utils.jl
```diff
@@ -17,6 +17,16 @@ y_train_regression = rand(1000)
 end
 
 
+@testset "stringifyparams -- multiple calls won't mutate fields" begin
```
kainkad, line 20: These tests also pass on the existing LightGBM without your change because the `indices` are being passed to `categorical_feature` so each time you call `stringifyparams` the values in the `indices` (assigned to `categorical_features`) will be also mutated, hence the test will always be true even without your change.

The scenario to be considered based on the #133 should be following:
```
categorical_features = [1, 3, 5, 7, 9]
expected_indices = [0, 2, 4, 6, 8] #as there's `-1` subtraction to match the C's zero-based indices.
classifier = LightGBM.LGBMClassification(categorical_feature = categorical_features)
LightGBM.stringifyparams(classifier; verbosity=-1)
@test expected_indices == classifier.categorical_feature

LightGBM.stringifyparams(classifier; verbosity=-1)
@test expected_indices == classifier.categorical_feature
```
With your current change the indices are not corrected and they remain [1, 3, 5, 7, 9] not [0, 2, 4, 6, 8] so this also needs to be revised.

This test sis better placed in `test_fit.jl` rather than utils as the `stringifparams` function is part of the `fit.jl` and the existing tests are in `test_fit.jl`

kainkad:
HI @characat0 thank you for raising the issue and submitting PR. I have left comments on your PR as it currently doesn't seem to address the issue raised. Are you ok to follow up on this?

characat0:
Thanks for the feedback, @kainkad. Will submit these changes.

characat0:
We missed a config for the test LGBM_BoosterUpdateOneIterCustom, in https://github.com/microsoft/LightGBM/pull/5438 they added an explicit test for objective==nullptr

characat0:
> I have a question on the null pointer, isn't it a default for the objective unless specified though?
I thought that too, but looks like the default objective is 'regression', I've confirmed this using LightGBM.LGBM_BoosterSaveModelToString

PR #135, opened 2023-05-11
`num_threads` for `predict`

artemsolod:
`predict` by default reuses `num_threads` from training. However LightGBM supports setting a separate `num_threads` for predict. Here is a python example https://github.com/microsoft/LightGBM/issues/4706.

I am not familiar with MLJ interface so I have added a `num_threads` keyword argument only to `predict` and `predict_classes`.

Also some of the test are failing but it appears to be unrelated to this PR (`KeyError: key "LIGHTGBM_EXAMPLES_PATH" not found`)

### DIFF
--- a/src/predict.jl
+++ b/src/predict.jl
@@ -17,6 +17,8 @@ Use `dropdims` if a vector is required.
     only, `0` includes warning logs, `1` includes info logs, and `> 1` includes debug logs.
 * `is_row_major::Bool`: keyword argument that indicates whether or not `X` is row-major. `true`
     indicates that it is row-major, `false` indicates that it is column-major (Julia's default).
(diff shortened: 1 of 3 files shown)


kainkad:
Hi @artemsolod thanks for the PR and adding num_threads to predict. The CI tests passed and the KeyError you observed is because locally you need to download and extract the LightGBM examples and then set the environment variable LIGHTGBM_EXAMPLES_PATH to the root of the source installation so it can use these examples for tests. It would be great to add a basic test for LGBM_BoosterPredictForMat but I'll leave it up to you. There's a small comment on the PR.

kainkad:
> Thanks for the comments @kainkad! I've added a simple test call for `LGBM_BoosterPredictForMat`, hope it's enough, as I see there are other tests on `predict` that use it internally.

Sounds good!

PR #136, opened 2023-05-19
add gpu platform, num_gpu, device_id to MLJ interface

kollienne:


kainkad:
Hi @kollienne thank you for adding these params. Looks good to me.  Just wanted to check whether you were able to test your branch and run with some gpus to see it works as expected via MLJ interface? In this case we can't really add a unit test to the repo as we don't have gpu testing workflow.
I'll wait for your answer but I'm expecting this should work fine as these are the same params we have in estimators that were tested with gpus back then so unless there's anything specific for MLJ we should consider, there shouldn't be any issue.

kainkad:
> It's working on my end with four gpus. The only thing I haven't tested is gpu_platform_id, since I only have one platform.

Great! I'm merging this PR in. I'll go through other PRs that can potentially go in and then hopefully should be able to release.

PR #152, opened 2024-07-24
Add lgbm ranker

### LINKED ISSUE #118: Feature Request: LightGBM Ranking
Hi, are there plans to add ranking to this package (LightGBM Ranker)? Currently the only ways to get ranking is going PyCall method for Catboost or convert all variables to numeric for XGBoost (this is a nightmare for large/wide datasets).

having a 100% Julia solution would be awesome.

kainkad:
-  closes #118:
  - by adding a separate ranking estimator 
  - adding group query info as a vector of ints that can be passed to the fit method for training data similarly to weight and input score data rather than through a separate `group_column` parameter requiring additional .txt files for query and training data [ref: query data](https://lightgbm.readthedocs.io/en/v3.3.5/Parameters.html#query-data)
-  matches with the provided python equivalent for fit and predictions (added to tests)

### DIFF
--- a/src/MLJInterface.jl
+++ b/src/MLJInterface.jl
@@ -83,7 +83,7 @@ MLJModelInterface.@mlj_model mutable struct LGBMRegressor <: MLJModelInterface.D
     metric::Vector{String} = ["l2"]::(all(in.(_, (LGBM_METRICS, ))))
     metric_freq::Int = 1::(_ > 0)
     is_training_metric::Bool = false
-    ndcg_at::Vector{Int} = Vector{Int}([1, 2, 3, 4, 5])::(all(_ .> 0))
+    eval_at::Vector{Int} = Vector{Int}([1, 2, 3, 4, 5])::(all(_ .> 0))
 
     # Implementation parameters
     num_machines::Int = 1::(_ > 0)
@@ -167,7 +167,7 @@ MLJModelInterface.@mlj_model mutable struct LGBMClassifier <: MLJModelInterface.
     metric::Vector{String} = ["None"]::(all(in.(_, (LGBM_METRICS, ))))
     metric_freq::Int = 1::(_ > 0)
     is_training_metric::Bool = false
-    ndcg_at::Vector{Int} = Vector{Int}([1, 2, 3, 4, 5])::(all(_ .> 0))
+    eval_at::Vector{Int} = Vector{Int}([1, 2, 3, 4, 5])::(all(_ .> 0))
 
     # Implementation parameters
     num_machines::Int = 1::(_ > 0)
--- a/src/estimators.jl
+++ b/src/estimators.jl
@@ -298,7 +298,7 @@ mutable struct LGBMClassification <: LGBMEstimator
     metric::Vector{String}
     metric_freq::Int
     is_training_metric::Bool
-    ndcg_at::Vector{Int}
+    ndcg_at::Vector{Int} #alias it's called eval_at
 
     num_machines::Int
     local_listen_port::Int
--- a/src/estimators.jl
+++ b/src/estimators.jl
@@ -561,7 +561,6 @@ mutable struct LGBMRanking <: LGBMEstimator
     force_col_wise::Bool
     force_row_wise::Bool
 
-    group_column::String
     lambdarank_truncation_level::Int
     lambdarank_norm::Bool
     label_gain::Vector{Int}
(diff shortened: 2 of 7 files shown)


src/MLJInterface.jl
```diff
@@ -83,7 +83,7 @@ MLJModelInterface.@mlj_model mutable struct LGBMRegressor <: MLJModelInterface.D
     metric::Vector{String} = ["l2"]::(all(in.(_, (LGBM_METRICS, ))))
     metric_freq::Int = 1::(_ > 0)
     is_training_metric::Bool = false
-    ndcg_at::Vector{Int} = Vector{Int}([1, 2, 3, 4, 5])::(all(_ .> 0))
```
kainkad, line 86: The changes here and in the following structs are due to the alias being used instead of the actual parameter name [reference]((https://lightgbm.readthedocs.io/en/v3.3.5/Parameters.html#eval_at) . The aliases are not supported in LightGBM.jl 
This parameter is only applicable to `ndcg` and `map` metrics which are used in learning to rank applications so I was hesitating whether to remove it completely from Regressors and Classifiers since I created a separate Ranking estimator. It was there for many years so removing would be breaking if someone used these estimators (and technically  eval_at still works with these estimators when the correct metric is used) so I decided to leave it.

src/estimators.jl
```diff
@@ -472,8 +473,267 @@ function LGBMClassification(;
...
         drop_rate, max_drop, skip_drop, xgboost_dart_mode,
         uniform_drop, drop_seed, top_rate, other_rate, min_data_per_group, max_cat_threshold, cat_l2, cat_smooth,
-        metric, metric_freq, is_training_metric, ndcg_at, num_machines, local_listen_port, time_out,
+        metric, metric_freq, is_training_metric, eval_at, num_machines, local_listen_port, time_out,
         machine_list_file, num_class, device_type, gpu_use_dp, gpu_platform_id, gpu_device_id, num_gpu,
         force_col_wise, force_row_wise,
     )
 end
+
+mutable struct LGBMRanking <: LGBMEstimator
```
kainkad, line 482: I decided to create a separate ranking estimator (which has majority of params shared with the classifier plus some added new one for `lambdarank` and `rank_xendcg` objectives) as there might be some future changes that would be only applicable to the ranking tasks such as adding evaluation set and evaluation group query ids/sizes to the fit method.  The default objective has been set to `lambdarank` . This estimator is currently only available in the core LightGBM.jl, it's not added to the MLJ interface. This can be done in the future once it proves to work well.

test/integration/groupQueryTest.jl
```diff
@@ -0,0 +1,106 @@
...
+    )
+
+    # Fit the model
+    LightGBM.fit!(ranker, X_train, Vector(y_train), group = group_train)
+
+    # Predict the relevance scores for the test set
+    y_pred = LightGBM.predict(ranker, X_test)
+
+    # Test the predicted scores are as expected to match with python's output
+    @test y_pred ≈ [0.3713922520492622; -0.15637136430479565] atol=1e-6
```
kainkad, line 101: I thought a matching result with python's output for predictions being the relevance scores rather than the predicted probabilities could be good to start with given such a small test data but may be some more sophisticated test to check whether the ordering of those predicted relevance scores matches the actual could be updated at some point.

src/estimators.jl
kainkad: I decided to bring the `group_column` parameter since it's available in LightGBM but it does require the model to be stored as .txt which is not the usual use case as described above.

src/fit.jl
```diff
@@ -1,5 +1,13 @@
 """
-    fit!(estimator, num_iterations, X, y[, test...]; [verbosity = 1, is_row_major = false])
```
kainkad, line 2: Just a comment here for reference. This change is just a removal  of an outdated doc signature for a method with iterations that's no longer used. The `num_iterations` has been removed from the `fit!` method in this [PR](https://github.com/IQVIA-ML/LightGBM.jl/pull/80/files) as it was not necessary (train! still has num_iterations). The `num_iterations` were initially introduced in the `fit!` but at the time it wasn't clear this was needed in the `fit!` method itself [reference](https://github.com/IQVIA-ML/LightGBM.jl/pull/51/files#r425872182).

BojieSheng (approved):
LGTM.

PR #153, opened 2024-08-15
added tests to DatasetSetField

kainkad:
- added tests to DataSetField

### DIFF
--- a/src/wrapper.jl
+++ b/src/wrapper.jl
@@ -319,24 +319,11 @@ function _LGBM_DatasetSetField(ds::Dataset, field_name::String,
     return nothing
 end
 
-function LGBM_DatasetSetField(ds::Dataset, field_name::String, field_data::Vector{Float32})
-    if field_name == "label" || field_name == "weight"
-        _LGBM_DatasetSetField(ds, field_name, field_data)
-    else
-        _LGBM_DatasetSetField(ds, field_name, convert(Vector{Int32}, field_data))
-    end
-    return nothing
-end
-
-function LGBM_DatasetSetField(ds::Dataset, field_name::String, field_data::Vector{Int32})
-    if field_name == "group"
-        _LGBM_DatasetSetField(ds, field_name, field_data)
-    else
-        _LGBM_DatasetSetField(ds, field_name, convert(Vector{Float32}, field_data))
-    end
-    return nothing
-end
-
+# the below type conversions are required by C API for the different fieldnames
+# ideally we should enforce the correct types through signature and multiple dispatch accordingly
+# but this would restrict the inputs from MLJInterface which don't necessarily have
+# exact matching data types for these fields hence leaving the conversions here before passing to the
+# wrapper function
 function LGBM_DatasetSetField(ds::Dataset, field_name::String, field_data::Vector{T}) where T<:Real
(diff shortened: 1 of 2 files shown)


src/wrapper.jl
```diff
@@ -319,24 +319,11 @@ function _LGBM_DatasetSetField(ds::Dataset, field_name::String,
...
-    end
-    return nothing
-end
-
+# the below type conversions are required by C API for the different fieldnames
+# ideally we should enforce the correct types through signature and multiple dispatch accordingly
+# but this would restrict the inputs from MLJInterface which don't necessarily have
+# exact matching data types for these fields hence leaving the conversions here before passing to the
+# wrapper function
 function LGBM_DatasetSetField(ds::Dataset, field_name::String, field_data::Vector{T}) where T<:Real
```
kainkad, line 327: ideally here we should have something like this (with or without the error handling and may be even with structs for field_name):

```
function LGBM_DatasetSetField(ds::Dataset, field_name::String, field_data::Vector{Float32})
    if field_name == "label" || field_name == "weight"
        _LGBM_DatasetSetField(ds, field_name, field_data)
    else
        error("Invalid field name: $field_name for Float32 data")
    end
    return nothing
end

function LGBM_DatasetSetField(ds::Dataset, field_name::String, field_data::Vector{Float64})
    if field_name == "init_score"
        _LGBM_DatasetSetField(ds, field_name, field_data)
    else
        error("Invalid field name: $field_name for Float64 data")
    end
    return nothing
end

function LGBM_DatasetSetField(ds::Dataset, field_name::String, field_data::Vector{Int32})
    if field_name == "group"
        _LGBM_DatasetSetField(ds, field_name, field_data)
    else
        error("Invalid field name: $field_name for Int32 data")
    end
    return nothing
end

```

but this seems too restrictive especially for labels coming from interfaces calling lgbm which don't have such type restrictions
```

File: test/ffi/datasets.jl [hunk 2]
```diff
@@ -144,9 +144,40 @@ end
 
 
 @testset "LGBM_DatasetSetField" begin
-    # The current implementation for LGBM_DatasetSetField has questionable type promotions
-    # but it touches too much code in order to tidy up right now
-    @test_broken false
+    mymat = [1. 2.; 3. 4.; 5. 6.]
+    ds = LightGBM.LGBM_DatasetCreateFromMat(mymat, verbosity)
+
+    @testset "_LGBM_DatasetSetField invalid field name test" begin
```
kainkad, line 150: This test is essentially checking the type conversions and field allocation which is what the function is doing.

src/wrapper.jl
```diff
@@ -319,24 +319,11 @@ function _LGBM_DatasetSetField(ds::Dataset, field_name::String,
     return nothing
 end
 
-function LGBM_DatasetSetField(ds::Dataset, field_name::String, field_data::Vector{Float32})
```
kainkad, line 322: I removed these as all the cases are handled by the conversions done in the remaining function. These didn't cover the case of "init_score" which is expected to be Float64 so they were redirecting between label, weight and group field names and using the last function for the init score, which also handles all other field names anyway.
Given my other comments on limitations of restricting the field_data type through signatures, the remaining function if fine to cover all cases.

Issue #23, opened 2020-02-19
consider how we can bring more lightgbm parameters


yalwan-iqvia:
https://lightgbm.readthedocs.io/en/latest/Parameters.html

yalwan-iqvia:
Defaults might not be matching main LightGBM: https://github.com/IQVIA-ML/LightGBM.jl/issues/42#issuecomment-614185868
Another vote for the `boosting` parameter: https://github.com/IQVIA-ML/LightGBM.jl/issues/42#issuecomment-614194236

@azev77

azev77:
Btw, Caret has an interface to many R packages, just like MLJ. 
In their interface, they set: xgbDART, xgbLinear, xgbTree as 3 separate models instead of just choosing the parameter xgb(boost="dart"), xgb(boost="linear") etc 

Might this be worth doing w/ your interface to MLJ?

yalwan-iqvia:
In implementation terms, doing this is likely to result in massive amounts of duplication of code.
I'm also not really sure of what the advantage is in real terms is to the user.

If you see recent commit https://github.com/IQVIA-ML/LightGBM.jl/commit/ee01161c0bb74cdf1cb47ccdefddc0e1886c475f that we actually went in opposite direction to this, by removing object-wise distinction between binary-multiclass (which just corresponds to a change in the value for `objective` parameter)

However, I agree that the missing support for the boosting parameter needs to be added.

azev77:
You’re probably right. 
The only reason I brought it up is that it can make it look like MLJ has more models. 
Carey claims to have > 200 models, but a lot of them are the same model w different HP 

xgbDART, xgbLinear, xgbTree

yalwan-iqvia:
Just to let you know, I haven't forgotten this. I've been buried with other work, but I'm hoping in the coming weeks to bundle a few features together and create a new release.

sbeura:
i found these important parameters missing:
 
**boosting**: defines the type of algorithm you want to run, default=gdbt
    gbdt: traditional Gradient Boosting Decision Tree
    rf: random forest
    dart: Dropouts meet Multiple Additive Regression Trees
    goss: Gradient-based One-Side Sampling
 
**max_cat_group**: When the number of category is large, finding the split point on it is easily over-fitting. So LightGBM merges them into ‘max_cat_group’ groups, and finds the split points on the group boundaries, default:64

[...]

PR #154, opened 2024-08-16
Update alias params

kainkad:
- updated params with their documented names not aliases
- update default values to match lightgbm [documentation](https://lightgbm.readthedocs.io/en/v3.3.5/Parameters.html)
- re-enabled formerly broken tests
- relates to https://github.com/IQVIA-ML/LightGBM.jl/issues/23#issuecomment-614446682 (but leaves the issue opened due to other discussions)


src/estimators.jl
```diff
@@ -83,14 +83,14 @@ end
     LGBMRegression(; [
         objective = "regression",
         boosting = "gbdt",
-        num_iterations = 10,
```
kainkad, line 86: in few cases the default parameter values differed to the official documentation defaults without any specific reason so I updated them accordingly given that all other parameters have defaults according to lightgbm documentation

src/estimators.jl
```diff
@@ -701,9 +701,9 @@ function LGBMRanking(;
     max_cat_threshold = 32,
     cat_l2 = 10.0,
     cat_smooth = 10.0,
-    metric = [""],
+    metric = ["ndcg"],
```
kainkad, line 704: Updated this default as the default objective for this estimator is lambdarank and these two usually go together. Unlike for binary, multiclass or regression objectives, where the empty "" metric is filled with default value corresponding to the pre-defined objective, the empty string doesn't assign the metric in this case.

src/estimators.jl
```diff
@@ -61,13 +60,13 @@ mutable struct LGBMRegression <: LGBMEstimator
...
+    is_provide_training_metric::Bool
     eval_at::Vector{Int}
 
     num_machines::Int
     local_listen_port::Int
     time_out::Int
-    machine_list_file::String
+    machine_list_filename::String
 
     num_class::Int
```
kainkad, line 71: `num_class` shouldn't be a parameter for the regressor but it's hardcoded to 1 and internally used to return reshaped predictions: https://github.com/IQVIA-ML/LightGBM.jl/blob/master/src/predict.jl#L41-L43 so removing it would require to find some decent equivalent in returning these predictions

src/MLJInterface.jl
```diff
@@ -82,15 +81,15 @@ MLJModelInterface.@mlj_model mutable struct LGBMRegressor <: MLJModelInterface.D
     # Metrics
     metric::Vector{String} = ["l2"]::(all(in.(_, (LGBM_METRICS, ))))
```
kainkad, line 82: Unfortunately defaulted values are required on the MLJInterface so here ["l2"] has been left where it is [""] in estimators.jl (which means a default for pre-defined objectives).
This applies to all the cases where we could use an empty string. 
image

src/estimators.jl
```diff
@@ -83,14 +82,14 @@ end
...
         boosting = "gbdt",
-        num_iterations = 10,
+        num_iterations = 100,
         learning_rate = .1,
-        num_leaves = 127,
+        num_leaves = 31,
         max_depth = -1,
         tree_learner = \"serial\",
-        num_threads = Sys.CPU_THREADS,
+        num_threads = 0,
```
kainkad, line 90: as per documentation the real cpu cores achieve better speed but the actual default is 0 so changed it across all https://lightgbm.readthedocs.io/en/v3.3.5/Parameters.html#num_threads

src/estimators.jl
```diff
@@ -371,14 +366,14 @@ end
         max_cat_threshold = 32,
         cat_l2 = 10.0,
         cat_smooth = 10.0,
-        metric = [\"multi_logloss\"],
+        metric = [""],
```
kainkad, line 369: Default for metric is an empty string `""` not a vector but this applies to one supported metric, or metrics introduced as a string such as `metric=binary_logloss` whereas for most use cases with multiple metrics we need a vector of strings, hence leaving it here.

src/MLJInterface.jl
```diff
@@ -155,7 +153,7 @@ MLJModelInterface.@mlj_model mutable struct LGBMClassifier <: MLJModelInterface.
     objective::String = "multiclass"::(_ in CLASSIFICATION_OBJECTIVES)
```
kainkad, line 153: Can't comment against the commented out sigmoid param hence commenting here (see lines above for sigmoid). I'm not quite sure about adding sigmoid here because of this: https://github.com/IQVIA-ML/LightGBM.jl/blob/master/src/MLJInterface.jl#L150-L152. Since the defaulted objective on the MLJ classifier is multiclass, then adding a defaulted sigmoid here might cause some issues but on the other hand the LGBMClassification also is defaulted to multiclass and still has the sigmoid defaulted to 1.

src/estimators.jl
```diff
@@ -371,14 +366,14 @@ end
...
         metric_freq = 1,
-        is_training_metric = false,
+        is_provide_training_metric = false,
         eval_at = Int[1, 2, 3, 4, 5],
         num_machines = 1,
         local_listen_port = 12400,
         time_out = 120,
-        machine_list_file = \"\",
+        machine_list_filename = \"\",
         num_class = 2,
```
kainkad, line 377: This does not match the default in the documentation but it' s because in the documentation also has the default objective set to regression for which num_class is also defaulted to 1.
In LighGBM.jl it's been decided to have now 3 estimators where the LGBMClassification is defaulted to multiclass (not binary) hence this default num_class.

src/MLJInterface.jl
kainkad: Re-enabled `sigmoid` with a default value since we have several other parameters which work only with `binary` or `multiclassova` such as `scale_pos_weight` or `is_unbalance`. Their defaults have no impact with the currently default objective set to `multiclass` but are made available if the user changes the objective and wants to use them.

src/estimators.jl
kainkad: Just to clarify on this the regression task does not use num_class which is required for classification task. The default in this estimator differs from the default in the official docs because the classifier in LightGBM.jl is defaulted to multiclass objective.

PR #159, opened 2024-09-12
added predict params except for prediction early stopping

kainkad:
- added prediction params except for early stopping which needs a separate change to support them

### DIFF
--- a/src/predict.jl
+++ b/src/predict.jl
@@ -1,4 +1,10 @@
 # TODO: Change verbosity on LightGBM's side after a booster is created.
+
+const C_API_PREDICT_NORMAL = 0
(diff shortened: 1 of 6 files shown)


src/predict.jl
```diff
@@ -32,6 +38,16 @@ function predict(
 
     tryload!(estimator)
 
+    if estimator.predict_raw_score && !estimator.predict_leaf_index && !estimator.predict_contrib
```
kainkad, line 41: I didn't include the same in the `predict_classes` function although it also takes the predict_type parameter that could be altered depending on the prediction params but basing on its functionality it is based on what is called the normal prediction (standard 0 .0 to 1.0 predicted probabilities)

PR #160, opened 2024-09-25
added pred early stopping and updated num_threads

kainkad:
- added prediction early stopping for Classification and Ranking
- updated `num_threads` to be read directly from the estimator rather than a separate param in predictions

### DIFF
--- a/src/estimators.jl
+++ b/src/estimators.jl
@@ -319,6 +319,9 @@ mutable struct LGBMClassification <: LGBMEstimator
     predict_leaf_index::Bool
     predict_contrib::Bool
     predict_disable_shape_check::Bool
+    pred_early_stop::Bool
+    pred_early_stop_freq::Int
+    pred_early_stop_margin::Float64
 
     # Objective parameters
     num_class::Int
@@ -403,6 +406,9 @@ end
         predict_leaf_index = false,
         predict_contrib = false,
         predict_disable_shape_check = false,
+        pred_early_stop = false,
+        pred_early_stop_freq = 10,
+        pred_early_stop_margin = 10.0,
         num_class = 2,
         is_unbalance = false,
         scale_pos_weight = 1.0,
@@ -480,6 +486,9 @@ function LGBMClassification(;
     predict_leaf_index = false,
     predict_contrib = false,
     predict_disable_shape_check = false,
+    pred_early_stop = false,
+    pred_early_stop_freq = 10,
+    pred_early_stop_margin = 10.0,
     num_class = 2,
     is_unbalance = false,
     scale_pos_weight = 1.0,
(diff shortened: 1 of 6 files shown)


src/estimators.jl
```diff
@@ -319,6 +319,9 @@ mutable struct LGBMClassification <: LGBMEstimator
     predict_leaf_index::Bool
     predict_contrib::Bool
     predict_disable_shape_check::Bool
+    pred_early_stop::Bool
```
kainkad, line 322: As per documentation prediction early stopping only applies to classification and ranking applications.

src/predict.jl
```diff
@@ -33,7 +31,6 @@ function predict(
     estimator::LGBMEstimator, X::AbstractMatrix{TX}; predict_type::Integer = 0,
     start_iteration::Integer = 0, num_iterations::Integer = -1, verbosity::Integer = 1,
     is_row_major::Bool = false,
-    num_threads::Integer = -1,
```
kainkad, line 36: I removed a separate num_threads parameter from `predict()` because it is a valid model parameter which can be taken directly from the model (either the default or what user decided to set up) without having to pass additionally pass it.  This way we could end up with quite a few params that could be added to predict() for e.g. the pred early stop or the predict type params but all of them can be read from the model.

src/wrapper.jl
```diff
@@ -665,10 +665,6 @@ function LGBM_BoosterPredictForMat(
     alloc_len = LGBM_BoosterCalcNumPredict(bst, nrow, predict_type, start_iteration, num_iteration)
     out_result = Array{Cdouble}(undef, alloc_len)
 
-    parameter = ""  # full prediction, no early stopping
```
kainkad, line 668: Given the previous comment this is not needed. `num_threads` is always specified as a param (either with a default or customised) and now the additionally there is an option for prediction early stopping.

test/basic/test_parameters.jl
```diff
@@ -298,4 +298,83 @@ end
 end
 
 
+@testset "parameters -- prediction with early stopping" begin
```
kainkad, line 301: Ideally, this test should show for e.g. that there is 100 iterations for full prediction and 30 for early predictions as   adding output from later trees is not improving the predictions.
I found it difficult to access such information through the test so instead I went for comparing the cases with and without early stop where with a very low margin for early prediction stopping, these predictions will be less accurate and not matching the full predictions.
For regression there is no prediction early stopping and those parameters cannot be provided so the test for that can be removed.

kainkad: Removed tests with regressor because they were not needed

src/predict.jl
kainkad: brough back `num_threads` in predictions as this allows to specify number of threads specifically for predictions which is separate to the number of threads used in training (which can be provided through `estimator.num_threads`

src/wrapper.jl
kainkad: Given the  second comment regarding number of threads, I've refactored this function and removed the `num_threads` as the number of threads will be informed through the `parameter` parameter.

src/predict.jl
```diff
@@ -50,8 +50,18 @@ function predict(
 
     log_debug(verbosity, "Started predicting\n")
 
+    if num_threads != -1
```
kainkad, line 53: I moved the logic for `num_threads` here and pass it together with parameters so there is no need to add an extra parameter to the wrapper for `LGBM_BoosterPredictForMat`. The initial condition was `if num_threads > 0` and only in such case this parameter was passed to LGBM_BoosterPredictForMat, otherwise it wasn't specified. I now made it so that it's always passed in parameters and either is what user provided in `predict()`or from estimator.

PR #161, opened 2024-10-07
Implement lgbm booster refit

kainkad:
- adds FFI for `LGBM_BoosterRefit`
- partially addresses #7 (as `refit_decay_rate` needs to be implemented as a part of a `refit()` function that uses `LGBM_BoosterRefit and `refit_decay_rate`` so this is bringing the FFI first and the default refit_decay_rate)

### DIFF
--- a/test/ffi/booster.jl
+++ b/test/ffi/booster.jl
@@ -414,9 +414,6 @@ end
     # Fit the estimator with the training data
     LightGBM.fit!(estimator, X_train, y_train, verbosity = -1)
     
-    # Get the number of trees in the booster
-    num_trees = LightGBM.LGBM_BoosterGetCurrentIteration(estimator.booster)
-    
     # Get the leaf predictions using the training data
     leaf_predictions = LightGBM.predict(estimator, X_train)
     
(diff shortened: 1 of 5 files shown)


test/ffi/booster.jl
```diff
@@ -403,6 +403,33 @@ end
...
+    
+    # Fit the estimator with the training data
+    LightGBM.fit!(estimator, X_train, y_train, verbosity = -1)
+    
+    # Get the leaf predictions using the training data
+    leaf_predictions = LightGBM.predict(estimator, X_train)
+
+    # Convert to Int32 which is expected by the C API and should be for leaf indices
+    # Reshape the leaf predictions as unlike the normal or raw predictions their number of rows is number of data points * num_trees/iterations 
+    reshaped_leaf_predictions = reshape(convert(Matrix{Int32}, leaf_predictions), (size(X_train, 1), estimator.num_iterations))
```
kainkad, line 422: This conversion can be probably done on predict level if parameter is predict_leaf_index because indices are not floats. As for reshaping it could be an optional param in predict. Reshaping can also be done prior to passing over to LGBM_BoosterRefit as that's where it's needed.

kainkad:
closing as included in #162

PR #164, opened 2024-10-18
Added LGBM_DatasetCreateFromFile

kainkad:


test/ffi/datasets.jl
```diff
@@ -74,6 +74,47 @@ end
...
+    end
+
+    # Define dataset parameters as strings
+    params = [
+        "two_round=true header=true label_column=0 weight_column=1 ignore_column=2",
+        "two_round=false header=true label_column=1 weight_column=2 ignore_column=3"
+    ]
+
+    expected_num_data = [3, 3]
+    expected_num_feature = [4, 4]
```
kainkad, line 97: need to update this to have difference between the params for e.g. with ignoring columns so the created datasets are different in the number of features

test/ffi/datasets.jl
```diff
@@ -74,6 +74,61 @@ end
...
+    0.1,0.2,0.3,0.4,0.5
+    0.6,0.7,0.8,0.9,1.0
+    1.1,1.2,1.3,1.4,1.5
+    """
+    sample_file = "sample_data.csv"
+    open(sample_file, "w") do f
+        write(f, sample_data)
+    end
+
+    # Define dataset parameters as strings
```
kainkad, line 90: There's some unclear behaviour for the `ignore_column` parameter. According to the docs it can be specified for the text files just like the `header` (which works) and the other params in this test: https://lightgbm.readthedocs.io/en/v3.3.5/Parameters.html#ignore_column.  Despite mentioning "text files", only CSV, TSV, and LibSVM (zero-based) see to be supported (https://github.com/microsoft/LightGBM/blob/master/src/io/parser.cpp#L264-L266) and the `sample_data` seems to be a valid example and other params do work with it.

PR #166, opened 2024-11-13
added verbosity param

kainkad:
- added verbosity param


src/cv.jl
```diff
@@ -24,22 +24,24 @@ last valid iteration.
...
 end
 
 # Pass Dataset class directly. This will speed up the process if it is part of an iterative process and a pre-constructed dataset is available
-function cv(estimator::LGBMEstimator, dataset::Dataset, splits; verbosity::Integer = 1, truncate_booster::Bool=true)
+function cv(estimator::LGBMEstimator, dataset::Dataset, splits; verbosity::Integer = nothing, truncate_booster::Bool=true)
 
     start_time = now()
     num_data = LGBM_DatasetGetNumData(dataset)
-    parameters = stringifyparams(estimator; verbosity=verbosity)
+    verbosity = isnothing(verbosity) ? estimator.verbosity : verbosity
```
kainkad, line 43: These particular changes where functions' verbosity was set to nothing are done for the functions which use stringifyparams to avoid double setting the verbosity (i.e. the default one on estimators + a default one with a specific integer value in the function) which causes a warning. This way the verbosity of the function, i.e. cv, fit! can still be handled via its parameters as before but won't conflict with the now defaulted param on estimator's level.

PR #167, opened 2024-11-29
added alignment for zero-based index for interaction_constraints

kainkad:
- added `interaction_constraints` to the `INDEXPARAM`
- updated type for `interaction_constraints` to accept now `Vector{Vector{Int}}` instead of String (default for CLI but supported as a list of list in other languages)
- aligned julia's one-based indexing to the zero-based indexing similar to `categorical_feature`

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -1,5 +1,5 @@
 
-const INDEXPARAMS = [:categorical_feature]
+const INDEXPARAMS = [:categorical_feature, :interaction_constraints]
 
 const MAXIMIZE_METRICS = ["auc", "ndcg", "average_precision"]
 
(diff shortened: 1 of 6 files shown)


src/utils.jl
```diff
@@ -1,5 +1,5 @@
 
-const INDEXPARAMS = [:categorical_feature]
+const INDEXPARAMS = [:categorical_feature, :interaction_constraints]
```
kainkad, line 2: Extended this to `interaction_constraints` param which requires similar zero-based indexing handling to `categorical_feature`.
There are some dataset params such as `label_column`, `weight_column`, or `group_column` can take a single index indicating the column, or the column name  and these are added in https://github.com/IQVIA-ML/LightGBM.jl/pull/164/files.

PR #172, opened 2025-02-12
extended fit and predict to support missing features in core lgbm

kainkad:
- adds conversion of julia missings to nan required in core lightgmb
- adds tests for `use_missing` and `zero_as_missing` params

### DIFF
--- a/src/predict.jl
+++ b/src/predict.jl
@@ -29,6 +29,20 @@ function get_predict_type(
     return predict_type
 end
 
+function convert_to_nan(mat::AbstractMatrix{Union{Missing, T}}) where {T<:Real}
+    # If the type is Int, cast to Float64
+    if T <: Int
+        mat = Float64.(mat)
+    end
+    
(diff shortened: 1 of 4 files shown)


src/predict.jl
```diff
@@ -29,6 +29,20 @@ function get_predict_type(
     return predict_type
 end
 
+function convert_to_nan(mat::AbstractMatrix{Union{Missing, T}}) where {T<:Real}
```
kainkad, line 32: This is a duplication of what's in the `fit.jl` `prepare_matrix` and would have been better for maintenance to have this in one place (utils may be) but because of the files not being modules it's not possible to import those functions, copying them by using the include filepath runs into import issues and because these are just sort of "helper" functions, it didn't make much sense to export them in the lightgbm module. May recheck whether reordering of fit and predict `include` could solve the import issue, or happy to check any other suggestions rather than leaving as it is.

kainkad: sorted

